PR #1, opened 2019-10-18
Add wasm API to use biscuit on other platform

NotBad4U:
This PR adds a wasm API to the biscuit API.
The wasm binding is built with [wasm_bindgen](https://github.com/rustwasm/wasm-bindgen).
You can generate an NPM package, compile and run the wasm tests with `wasm-pack`.
The `tests/wasm_test.js` shows how you can use biscuit with wasm in a JS.

The PR is ready for a first review.

:warning: We have some `expect()` in the wasm code due to the use of `JsValue::from_serde`, `into_serde()`. This errors report a `Serde::Error` which doesn't implement the Traits `FromWasm` and `IntoWasm`. How would I treat this subject ? I propose to transform the `Serde::Error` in a JSON.


NotBad4U:
We don't have a lot of TU for the wasm right now. I'm just waiting for the rework on the biscuit API before adding more.

NotBad4U:
The wasm API has moved in it's own crate https://github.com/NotBad4U/biscuit-wasm to improve the compile time of this crate.

Geal:
not applicable anymore, the wasm crate does not need these modifications to the base crate

PR #2, opened 2019-10-22
Update the code example of the biscuit API in the README

NotBad4U:
The code in the README was out to date.
This PR updates this code.

Remark:
In the previous version of the API, the verifier was create by calling:
```rust
let mut v1 = Verifier::new();
...
v1.verify(&biscuit);
```
In the last version, the `Verifier::new` method is only public at the crate scope level: `pub(crate)`.
So we have to use the `biscuit.verify()` method to create a `Verifier`, which implies to recreate all the verifier. This seems to double the `verifier` part code. Should we defer this change?

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,7 @@ Non goals:
 In this example we will see how we can create a token, add some caveats, serialize and deserialize a token, append more caveats, and validate those caveats in the context of a request:
 
 ```rust
-extern crate rand;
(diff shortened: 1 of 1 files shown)


Geal:
it is fine to double the code here, since it is there to demonstrate the difference between a token and its attenuated version. In the common case we will only create one verifier for one token

PR #9, opened 2020-06-17
update prost

Keruspe:
Tests fail, but the same tests fail even without this patch too here


Keruspe:
@Geal I also have another one for the rustcrypto stuff but it will have to wait for https://github.com/dalek-cryptography/curve25519-dalek/pull/327 to get merged before

PR #10, opened 2020-06-18
impl TryFrom<&str> for Caveat

Keruspe:


Geal:
it should use the caveat parser, as that one can handle caveats separated by `||`: https://github.com/CleverCloud/biscuit-rust/pull/11/files#diff-2c09afcdc3c420ab0678ba9b5e83959cR97

PR #14, opened 2020-11-19
fix PublicKey::from_bytes signature to be consistent with PrivateKey

meh:
As discussed on IRC.


meh:
In lieu of ergonomics I think it makes more sense to take `&[u8]` from both, both are marked as fallible and this will allow using stuff like [secrets](https://docs.rs/secrets) in the future.

PR #32, opened 2021-02-14
fix: compilation errors in bitvec / funty…

agueganno:
It seems a transitive dependency of the project broke - bitvec - because
one of its own dependencies - funty - declared a global BITS const… hence breaking
the current project.

According to the bitvec project, a fix may be to force funty’s version
to 1.1.0.

The matching issue in bitvec :

https://github.com/bitvecto-rs/bitvec/issues/105

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ zeroize = { version = "1", default-features = false }
 thiserror = "1"
 rand = { version = "0.7" }
 inline-c = { version = "0.1", optional = true }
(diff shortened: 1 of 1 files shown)


agueganno:
This may not affect the 1.0 branch, though.

Geal:
it is fixed directly in nom, you just need to do a `cargo update` and you'll get nom 6.1.1

agueganno:
Thx.

PR #33, opened 2021-04-19
add examples of failing check parser

Keruspe:
Currently the Check parser doesn't always consume all of its data and we end up with incomplete checks.

This adds three examples of failing parsing. Not sure if the three are relevant, but if we don't support those three, the ones we don't sure should throw a parsing error.


Keruspe:
Actually there's a missing ( which confuses the parser here (or an extra ) depending on how you see it)

PR #36, opened 2021-07-23
Various parser improvements

divarvel:
- fix a date parsing bug present in `parse_source` and `parse_source_block`
- tests improvements


Geal:
thanks

PR #37, opened 2021-07-23
Derive `std::error::Error` on `parser::Error`

divarvel:


divarvel:
Looks like i have broken something since yesterday evening :thinking:

PR #40, opened 2021-09-09
Make code easier to read

RemiBardon:
- Rename `Symbol` by `SymbolIndex`
- Refactor `variables_set` into a `Rule` instance method
- Use `Rule::apply` in `Rule::find_match`
- Use `p.ids[index] = val.clone()` instead of uselessly store the value
- Rename `SymbolTable::get_s` by `SymbolTable::get_symbol`

### DIFF
--- a/src/datalog/mod.rs
+++ b/src/datalog/mod.rs
@@ -5,7 +5,6 @@ use std::convert::AsRef;
 use std::fmt;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-pub type Symbol = u64;
 mod expression;
 mod symbol;
 pub use expression::*;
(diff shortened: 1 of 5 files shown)


src/format/mod.rs
```diff
@@ -186,54 +183,34 @@ impl SerializedBiscuit {
...
             data: v,
             next_key: next_keypair.public(),
             signature: signature,
         });
 
-        Ok(t)
+        Ok(SerializedBiscuit {
+            root_key_id: self.root_key_id,
+            authority: self.authority.clone(),
+            blocks: blocks.clone(),
```
Geal, line 199: that last clone is not needed

src/datalog/mod.rs
```diff
@@ -101,80 +100,48 @@ impl fmt::Display for Fact {
...
-                    },
-                    _ => continue,
-                };
-
-                p.ids[index] = value.clone();
-            }
-
-            Some(Fact { predicate: p })
-        });
+        let mut it = self.apply(facts, symbols);
```
Geal, line 144: nice, I like it when code is deleted

PR #44, opened 2021-12-14
fix error formatters

divarvel:
thiserror format strings use a rust-like syntax, not a C-like syntax.

(thankfully this is not a _bad_ formatstring issue)


Geal:
Thanks

PR #52, opened 2022-01-28
Extend the example and add (hidden) assertions for query, fix world not being populated from token when querying.

MidasLamb:
Expand the example in the docs for querying an authorizer for a biscuit.

### DIFF
--- a/src/token/authorizer.rs
+++ b/src/token/authorizer.rs
@@ -267,12 +267,42 @@ impl<'t> Authorizer<'t> {
     {
         let rule = rule.try_into()?;
 
+        if let Some(token) = self.token.as_ref() {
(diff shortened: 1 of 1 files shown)


Geal:
ah, sorry, I committed https://github.com/biscuit-auth/biscuit-rust/commit/f3ce50ebaa9847153c65a9e32a025e78180305cc before seeing your PR

adding the facts in `add_token` is a bit simpler than adding them and removing them in query, especially considering we might want to query the entire world after authorize() was called.

Still, could you rebase your PR over the current master? Because the tests and example will be useful

MidasLamb:
I've first considered that, but I saw that in some other places the facts were cleared again, so I copied from those implementations. Rebased in the meantime

Geal:
thanks!

PR #57, opened 2022-03-11
Fix default symbol table

divarvel:
The default symbol table defined in the spec lists `time`, not `current_time`. This is not a big issue for tokens created, attenuated and verified by the same implementation, but it can cause issue for cross implementation use.

The Haskell impl follows the spec, I'm not sure about go and java.


divarvel:
So, given the fact that every implementation derived from the rust one has the same error, I think it'd be better to update the spec and the Haskell impl to match current use.

PR #63, opened 2022-04-24
remove leftover debugging println!

meh:
Surprised me a bit in the logs after upgrading :panda_face:


Geal:
oops sorry about that :sweat_smile:

PR #68, opened 2022-05-16
remove symbol leftovers

divarvel:
Symbol syntax was still accepted by the parser and used in tests.

I've removed support for symbols from the parser and updated all tests and examples.

The only remaining use of a symbol is in the description of test case #19, which is supposed to check restrictions for privileged symbols. Since they are not there anymore, idk what the test actually checks. Maybe it could just be dropped?

### DIFF
--- a/tests/rights.rs
+++ b/tests/rights.rs
@@ -12,11 +12,17 @@ fn main() {
 
     let mut builder = Biscuit::builder(&root);
 
-    builder.add_authority_fact(fact("right", &[s("authority"), string("file1"), s("read")]));
-    builder.add_authority_fact(fact("right", &[s("authority"), string("file2"), s("read")]));
     builder.add_authority_fact(fact(
         "right",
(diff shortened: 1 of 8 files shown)


tests/rights.rs
```diff
@@ -12,11 +12,17 @@ fn main() {
 
     let mut builder = Biscuit::builder(&root);
 
-    builder.add_authority_fact(fact("right", &[s("authority"), string("file1"), s("read")]));
-    builder.add_authority_fact(fact("right", &[s("authority"), string("file2"), s("read")]));
     builder.add_authority_fact(fact(
         "right",
-        &[s("authority"), string("file1"), s("write")],
+        &[string("authority"), string("file1"), string("read")],
```
Geal, line 17: authority can be removed here no?

Geal (approved):

tests/rights.rs
divarvel: I don't think nullary facts are allowed

PR #73, opened 2022-05-18
Fix deprecated `ed25519_dalek::Signature::new` calls.

ptondereau:
In the next release of `ed25519_dalek` signature's[ parsing could fail.](https://github.com/RustCrypto/signatures/pull/401)
I've created  specific format errors but I can revert to `?` syntax instead

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -131,6 +131,10 @@ pub enum Format {
     InvalidSignatureSize(usize),
     #[error("invalid key")]
     InvalidKey(String),
+    #[error("could not deserialize signature")]
+    SignatureDeserializationError(String),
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -131,6 +131,10 @@ pub enum Format {
     InvalidSignatureSize(usize),
     #[error("invalid key")]
     InvalidKey(String),
+    #[error("could not deserialize signature")]
+    SignatureDeserializationError(String),
+    #[error("could not deserialize the signature's block")]
```
divarvel, line 136:
```suggestion
    #[error("could not deserialize the block signature")]
```

PR #76, opened 2022-05-19
Add macros for compile-time datalog parsing

divarvel:
The gist is to allow this kind of code:

```rust
 extern crate biscuit_auth;
 extern crate biscuit_quote;
 use biscuit_auth::KeyPair;
 use biscuit_quote::{authorizer, biscuit, block};
 use std::time::{Duration, SystemTime};

 let root = KeyPair::new();

 let biscuit = biscuit!(
   &root,
   r#"
   user({user_id});
   right({user_id}, "file1", "read");
   "#,
   user_id = "1234",
 ).build().expect("Failed to create biscuit");

 biscuit.append(block!(
   r#"
     check if time($time), $time < {expiration};
   "#,
   expiration = SystemTime::now() + Duration::from_secs(86_400),
 )).expect("Failed to append block");

 biscuit.authorize(&authorizer!(
   r#"
      time({now});
      operation({operation});
      resource({resource});

      is_allowed($user_id) <- right($user_id, $resource, $operation),
                              resource($resource),
                              operation($operation);

      allow if is_allowed({user_id});
   "#,
   now = SystemTime::now(),
   operation = "read",
   resource = "file1",
   user_id = "1234",
 )).expect("Failed to authorize biscuit");
```

# Remaining improvements

The happy path is good, but the error handling is rather bad for now. 

- macro errors could be better formatted
- parameter substitution errors should be collected instead of crashing at the first one

I'm not super familiar with reëxports in rust, but if we could have a single crate that exports the macros and `biscuit-rust`, that would be better in terms of ux (a bit like what clap now does, instead of requiring people to depend on two separate crates)

### DIFF
--- a/biscuit-auth-datalog-macros/src/lib.rs
+++ b/biscuit-auth-datalog-macros/src/lib.rs
@@ -1,5 +1,9 @@
 extern crate proc_macro;
-use biscuit_auth::{builder::BlockBuilder, error, parser::parse_block_source};
+use biscuit_auth::{
+    builder::{BlockBuilder, Check, Fact, Policy, Rule},
+    error,
+    parser::{parse_block_source, parse_source},
+};
 use proc_macro::TokenStream;
 use quote::{quote, ToTokens};
 use std::collections::HashMap;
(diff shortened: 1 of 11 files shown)


biscuit-auth-datalog-macros/src/lib.rs
```diff
@@ -0,0 +1,466 @@
...
+}
+
+impl ToTokens for BlockBuilderWithParams {
+    fn to_tokens(&self, tokens: &mut quote::__private::TokenStream) {
+        let param_names: Vec<String> = self.parameters.clone().into_keys().collect();
+        let param_values: Vec<Expr> = self.parameters.clone().into_values().collect();
+        let facts_quote = self.builder.facts.iter().map(|f| {
+            quote! {
+                let mut fact = #f;
+                #(fact.set_lenient(#param_names, #param_values).unwrap();)*
```
Geal, line 165: could the parameter names be checked at compilation time instead of using `set_lenient`?

biscuit-auth-datalog-macros/tests/simple_test.rs
```diff
@@ -0,0 +1,96 @@
...
+    let b = block!(
+        r#"fact("test", hex:aabbcc, [true], {my_key});
+            rule($0, true) <- fact($0, $1, $2, {my_key});
+            check if {my_key}.starts_with("my");
+            "#,
+        my_key = "my_value",
+        other_key = false,
+    );
+    assert_eq!(
+        format!("{}", &b),
```
Geal, line 17:
```suggestion
        format!("{}", &b),
```
```suggestion
        b.to_string(),
```

If `Display` is implemented, you get `ToString` too

Geal, line 17:
```suggestion
        format!("{}", &b),
```
```suggestion
        b.to_string(),
```

If `Display` is implemented, you get `ToString` too

biscuit-auth/src/token/builder.rs
```diff
@@ -594,6 +648,20 @@ impl fmt::Display for Predicate {
     }
 }
 
+#[cfg(feature = "datalog-macro")]
+impl ToTokens for Predicate {
+    fn to_tokens(&self, tokens: &mut quote::__private::TokenStream) {
+        let name = self.name.clone();
```
Geal, line 654: probably not necessary to clone here

Geal, line 654: probably not necessary to clone here

Geal (approved):

biscuit-auth-datalog-macros/src/lib.rs
divarvel: Yes, that's what I planned to do next. The PR is already big, so I'll do it in a follow-up PR

divarvel: to elaborate a bit:

- I'll aggregate all the parameter names from the datalog fragment and issue an error if it does not match exactly the provided map
- I think I will keep `set_lenient` to keep the generated code as simple as possible. the other solutions are to either:
  - filter the parameters within the quote (so this would happen at runtime, not very useful)
  - handle the filtering outside the quote, but that would mean having to slice in a much more complex data structure than a pair of vecs

meh:
Does this mean I can finally get rid of all the builders? :D

Also does this support the fancy interpolation stuff that recently landed on stable?

divarvel:
The goal definitely to get rid of manual builder stuff, yes.
It supports fancy interpolation (as showcased in the example of the PR desc), but that does not directly use `add_code_with_params`, due to how things are done. One subtle difference is that with the macro, you don't have to provide `Term` values directly, only `Into<Term>`, so that removes some noise.

The plan is to improve compile-time error reporting, and to expose everything from a single create, before releasing

meh:
Lovely, I've been really yearning for this :panda_face:

PR #82, opened 2022-07-28
Fix clippy warnings

matze:
A few of them point to unnecessary memory allocations.

### DIFF
--- a/biscuit-auth/src/datalog/mod.rs
+++ b/biscuit-auth/src/datalog/mod.rs
@@ -94,7 +94,7 @@ impl AsRef<Expression> for Expression {
     }
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Check {
     pub queries: Vec<Rule>,
 }
@@ -322,7 +322,7 @@ where
                 Some(variables) => {
                     //println!("predicates empty, will test variables: {:?}", variables);
                     let mut valid = true;
-                    let mut temporary_symbols = TemporarySymbolTable::new(&self.symbols);
+                    let mut temporary_symbols = TemporarySymbolTable::new(self.symbols);
                     for e in self.expressions.iter() {
                         match e.evaluate(&variables, &mut temporary_symbols) {
                             Some(Term::Bool(true)) => {}
@@ -386,7 +386,7 @@ where
                                     //println!("will test with variables: {:?}", variables);
                                     let mut valid = true;
                                     let mut temporary_symbols =
-                                        TemporarySymbolTable::new(&self.symbols);
+                                        TemporarySymbolTable::new(self.symbols);
                                     for e in self.expressions.iter() {
                                         match e.evaluate(&variables, &mut temporary_symbols) {
                                             Some(Term::Bool(true)) => {
@@ -443,7 +443,7 @@ where
     }
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct MatchedVariables {
     pub variables: HashMap<u32, Option<Term>>,
 }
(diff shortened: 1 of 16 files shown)


matze:
Considering the domain and security impact of the crate, it'd be very helpful that someone sets up Github actions workflow and checks with clippy on each push.

biscuit-auth/src/token/authorizer.rs
```diff
@@ -903,16 +903,16 @@ impl<'t> Authorizer<'t> {
         let (facts, rules, checks, policies) = self.dump();
         let mut f = String::new();
         for fact in facts {
-            f.push_str(&format!("{};\n", &fact));
+            let _ = writeln!(f, "{fact};");
```
Geal, line 906: nice :)

biscuit-auth/src/datalog/mod.rs
```diff
@@ -707,6 +706,12 @@ impl FactSet {
         self.inner.values().fold(0, |acc, set| acc + set.len())
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.inner
+            .values()
+            .fold(true, |acc, set| acc | set.is_empty())
```
Geal, line 712:
```suggestion
            .fold(true, |acc, set| acc && set.is_empty())
```

biscuit-auth/src/datalog/mod.rs
```diff
@@ -707,6 +706,12 @@ impl FactSet {
         self.inner.values().fold(0, |acc, set| acc + set.len())
     }
 
+    pub fn is_empty(&self) -> bool {
```
Geal, line 709: why add this method? It is not used anywhere

matze: Because clippy is … pedantic. It says, whenever there is a public `len()` method there should be a public `is_empty()` method because it might be to compute [less costly](https://rust-lang.github.io/rust-clippy/master/#len_without_is_empty). But I can also add an ignore if you prefer.

biscuit-auth/src/datalog/mod.rs
matze: Haha, clippy has yet another suggestion
```rust
self.inner.values().all(|set| set.is_empty())
```
but will drop it if you decide to ignore the warning.

matze:
P.S.: the recent large change caused this one here:
```
error: you are implementing `Hash` explicitly but have derived `PartialEq`
   --> biscuit-auth/src/crypto/mod.rs:147:1
    |
147 | / impl Hash for PublicKey {
148 | |     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
149 | |         (crate::format::schema::public_key::Algorithm::Ed25519 as i32).hash(state);
150 | |         self.0.to_bytes().hash(state);
151 | |     }
152 | | }
    | |_^
    |
    = note: `#[deny(clippy::derive_hash_xor_eq)]` on by default
note: `PartialEq` implemented here
   --> biscuit-auth/src/crypto/mod.rs:107:30
    |
107 | #[derive(Debug, Clone, Copy, PartialEq, Eq)]
    |                              ^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq
    = note: this error originates in the derive macro `PartialEq` (in Nightly builds, run with -Z macro-backtrace for more info)
```
since this is critical crypto code, I did not want to touch it.

biscuit-auth/src/datalog/mod.rs
Geal: right, that method is not useful right now

Geal:
> P.S.: the recent large change caused this one here:
> [...]
> since this is critical crypto code, I did not want to touch it.

yup, that one should not be modified because `Hash` cannot actually be derived from that type.

matze:
> yup, that one should not be modified because Hash cannot actually be derived from that type.

What clippy is trying to say is that one should also implement `PartialEq` manually to ensure `k1 == k2 ⇒ hash(k1) == hash(k2)`.

Geal:
alright, it's merged! sorry it took so long, there were a lot of API changes we needed to ship

PR #86, opened 2022-08-07
rewrite BiscuitBuilder to use BlockBuilder internally

### LINKED ISSUE #80: pass the root key and symbol table as arguments to BiscuitBuilder
this will simplify the API and the integration in wasm and Python

Geal:
fix #80


divarvel:
❤️🎉

PR #88, opened 2022-08-12
Reorganize crates

Geal:
* parsers are now extracted into biscuit-parser, with minimal builder-like structures to represent datalog
* biscuit-quote uses biscuit-parser
* biscuit-auth reexports biscuit-parser in its own parser module
* biscuit-auth reexports proc macros from biscuit-quote in its own macros module


divarvel:
🎉

PR #90, opened 2022-08-13
set block version to 4 for recent features

Geal:
blocks that use scopes and 3d party blocks should have their version set to 4, but we can keep version 3 if it does not use those features


divarvel (approved):
Neat! I wonder if the constants are still useful since the logic is now tied to specific versions, not just MinSBV / MaxSBV

PR #92, opened 2022-08-14
properly translate public keys when adding a token

Geal:
if the authorizer already has checks r rules using external keys, the
key indexes might be wrong when adding a token, so theyhould be
translated like symbols


divarvel (approved):
:+1:

PR #100, opened 2022-08-18
Use rule scopes in authorizer query

divarvel:


biscuit-auth/src/token/authorizer.rs
```diff
@@ -464,20 +463,26 @@ impl<'t> Authorizer<'t> {
...
+        let rule = rule.try_into()?.convert(&mut self.symbols);
+
+        let mut origin = rule.origins(usize::MAX, Some(&self.public_key_to_block_id));
+
+        // by default, the query trusts the authority, in addition to the authorizer itself
+        if rule.scopes.is_empty() {
+            origin.insert(0);
+        }
+
+        println!("{:?}", &origin);
```
Geal, line 480:
```suggestion
```

biscuit-auth/src/token/authorizer.rs
```diff
@@ -538,20 +543,25 @@ impl<'t> Authorizer<'t> {
...
-        let rule = rule.try_into()?;
+        let rule = rule.try_into()?.convert(&mut self.symbols);
 
         self.world
             .run_with_limits(&self.symbols, limits.into())
             .map_err(error::Token::RunLimit)?;
-        let rule = rule.convert(&mut self.symbols);
-        let origin = if let Some(t) = self.token {
+        let all_origins = if let Some(t) = self.token {
             std::iter::repeat(())
```
Geal, line 552: this should also contain the authorizer's scope no?

biscuit-auth/src/token/authorizer.rs
```diff
@@ -538,20 +543,25 @@ impl<'t> Authorizer<'t> {
...
                 .collect()
         } else {
-            [0].iter().collect()
+            [0, usize::MAX].iter().collect()
+        };
+
+        let origin = if rule.scopes.is_empty() {
+            all_origins
+        } else {
+            rule.origins(usize::MAX, Some(&self.public_key_to_block_id))
```
Geal, line 564: if the rule has a `previous` scope then this will generate a list of scopes from 0 to usize::MAX

Geal (changes requested):

biscuit-auth/src/token/authorizer.rs
divarvel: We should disallow trusting previous altogether in the authorizer imo, it's too dangerous

biscuit-auth/src/token/authorizer.rs
divarvel: It could be argued either way it think. Both are fine to me.

divarvel: If we see the query as a way to explore the final state of the world, compared to the authorizer whose role is to authorize or not the token, then not reusing the authorizer scope makes sense

Geal: I think it should have access to the authorizer too. Qury should beable to explore the entire space, and the authorizer facts are part of it

biscuit-auth/src/token/authorizer.rs
Geal: I agree we should remove it. If someone wants to use it anyway it's always possible to explicitely set the list of scopes

biscuit-auth/src/token/authorizer.rs
divarvel: Ah i might have misunderstood, i was talking about using a default scope annotation

PR #103, opened 2022-08-19
use a BlockBuilder in Authorizer

divarvel:
Instead of adding authorizer checks, rules & facts directly to the datalog world, we
keep them in a BlockBuilder, and add them to the world only when running `authorize`.

This brings ~two~ three important benefits:

- this removes duplication between authorizer and BlockBuilder by letting Authorizer
  delegate to BlockBuilder (`add_xxx`)
- this solves an issue where rules added in the authorizer before the token would not
  have access to the public_key_to_block_id mapping and would fail to generate facts
  for rules trusting third-party blocks
- this allows an authorizer to carry a default scope, like blocks do

A related refactoring has been included: instead of having `Block.origins()` and
`Rule.origins()` with duplicate logic, everything happens inside a single
`TrustedOrigins::from_scope` function that handles both the defaulting behaviour for scopes, and then the conversion
to trusted origins. Furthermore, a `TrustedOrigins` newtype has been introduced, to separate the origin of a fact,
from the origins trusted by a rule. This makes many previous bugs type errors.

Do note:

1. `add_code` in `BlockBuilder` does not handle block-leve scopes, and `add_code` in `Authorizer` doesn't eithe
2. `parse_source` refuses parsing a block-level scope annotation anyway
3. `query` called before `authorize` will only return token facts (only those declared directly, not those generated by rules).

1 & 2 will be covered in a dedicated PR (along with parameter substitution in macros)

3 will require further discussion around making `Authorizer` more type-safe, by making its lifecycle more obvious (currently a lot depends on when some methods are called)

### DIFF
--- a/biscuit-auth/src/token/authorizer.rs
+++ b/biscuit-auth/src/token/authorizer.rs
@@ -1,15 +1,17 @@
 //! Authorizer structure and associated functions
 use super::builder::{
-    constrained_rule, date, fact, pred, rule, string, var, Binary, Check, Expression, Fact, Op,
-    Policy, PolicyKind, Rule, Term,
+    constrained_rule, date, fact, pred, rule, string, var, Binary, BlockBuilder, Check, Expression,
+    Fact, Op, Policy, PolicyKind, Rule, Scope, Term,
 };
 use super::builder_ext::{AuthorizerExt, BuilderExt};
 use super::{Biscuit, Block};
 use crate::builder::Convert;
 use crate::crypto::PublicKey;
-use crate::datalog::{self, FactSet, Origin, RuleSet, RunLimits};
+use crate::datalog::{self, Origin, RunLimits};
 use crate::error;
 use crate::time::Instant;
+use crate::token;
+use crate::token::scope_to_origins;
 use biscuit_parser::parser::parse_source;
 use prost::Message;
 use std::collections::{BTreeMap, HashSet};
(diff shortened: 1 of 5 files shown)


biscuit-auth/src/token/authorizer.rs
```diff
@@ -143,10 +133,16 @@ impl<'t> Authorizer<'t> {
         let mut blocks = Vec::new();
 
         let authority = token.block(0)?;
-        let block_scope = authority.origins(0, Some(&self.public_key_to_block_id));
 
         let mut authority_origin = Origin::default();
         authority_origin.insert(0);
+
+        let authority_trusted_oringins = TrustedOrigins::from_scopes(
```
Geal, line 140:
```suggestion
        let authority_trusted_origins = TrustedOrigins::from_scopes(
```

biscuit-auth/src/token/authorizer.rs
```diff
@@ -914,29 +947,33 @@ impl<'t> Authorizer<'t> {
...
-            checks,
-            self.policies.clone(),
-        )
+        let mut facts = self
+            .world
+            .facts
+            .iter_all()
+            .map(|f| Fact::convert_from(f.1, &self.symbols))
+            .collect::<Result<Vec<_>, error::Format>>()
+            .unwrap();
```
Geal, line 964: We might want to return a Vec<Result> knqtead: if there was an invalid fact for any reason, we would like to see the valid facts and the errors

Geal (approved):

biscuit-auth/src/token/authorizer.rs
divarvel: gnnn, i made all the patch telling myself "don't forget to fix the typo when you're done"

biscuit-auth/src/token/authorizer.rs
divarvel: yeah. for authorization, we want an error to blow up the whole computation, but for querying i guess i can see the merit in ignoring errors.

divarvel: Let's decide this in the "revamp the authorizer lifecycle", since it will touch queries as well.

PR #107, opened 2022-09-27
Check all implementation

Geal:
experimental code for https://github.com/biscuit-auth/biscuit/issues/109


divarvel (commented):
This looks good! I've not dug into the iterator changes in the datalog engine.

One things that's missing though (i think) is a version check on serialization / deserialization, to make sure that only v4 biscuits can carry such a check (else an old implementation could ignore the added enum and treat a `check all` as a `check if`.

A third test could be added in token/mod.rs for a successful `check all` case, and also maybe a test with no bound variables, just performing an exact match (that's a corner case, but it's always good to pin this kind of behaviour)

PR #108, opened 2022-10-11
add support for ecdsa signatures

Geal:
* https://github.com/biscuit-auth/biscuit/pull/165


biscuit-auth/benches/token.rs
```diff
@@ -11,7 +11,7 @@ use rand::rngs::OsRng;
 
 fn create_block_1(b: &mut Bencher) {
     let mut rng = OsRng;
-    let root = KeyPair::new_with_rng(&mut rng);
+    let root = KeyPair::new_with_rng(Algorithm::Ed25519, &mut rng);
```
divarvel, line 14: it could be done in another PR, but adding benches for other signature algorithms would be nice (eg making the bench suite parametric on the algorithm, and running it for each supported algorithm)

biscuit-auth/src/crypto/mod.rs
```diff
@@ -176,55 +247,68 @@ impl PublicKey {
...
+            },
+        )?)
     }
 }
 
 impl Display for PublicKey {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "ed25519/{}", hex::encode(self.to_bytes()))
+        match self {
+            PublicKey::Ed25519(key) => write!(f, "ed25519/{}", hex::encode(key.to_bytes())),
```
divarvel, line 287: i think there should be a dedicated function for datalog rendering. Using `Display` here to generate datalog code is a bit fragile.
Also, both public key formats have a `print()` method that could be used to avoid duplication (if used for datalog formatting, it should also be named accordingly

biscuit-auth/src/datalog/symbol.rs
```diff
@@ -282,7 +282,7 @@ impl SymbolTable {
                     crate::token::Scope::Previous => "previous".to_string(),
                     crate::token::Scope::PublicKey(key_id) => {
                         match self.public_keys.get_key(*key_id) {
-                            Some(key) => format!("ed25519/{}", hex::encode(key.to_bytes())),
+                            Some(key) => key.to_string(),
```
divarvel, line 285: see comments in the `crypto` mod, I think it is a bit fragile to use `Display` for this

biscuit-auth/src/token/builder.rs
```diff
@@ -744,7 +754,10 @@ impl fmt::Display for Scope {
         match self {
             Scope::Authority => write!(f, "authority"),
             Scope::Previous => write!(f, "previous"),
-            Scope::PublicKey(pk) => write!(f, "ed25519/{}", hex::encode(pk.to_bytes())),
+            Scope::PublicKey(pk) => match pk {
+                PublicKey::Ed25519(key) => write!(f, "ed25519/{}", hex::encode(key.to_bytes())),
```
divarvel, line 758: see comment in `datalog/symbol.rs`

all this should be defined in a central location

biscuit-auth/src/token/builder.rs
```diff
@@ -757,15 +770,42 @@ impl From<biscuit_parser::builder::Scope> for Scope {
...
             biscuit_parser::builder::Scope::Authority => Scope::Authority,
             biscuit_parser::builder::Scope::Previous => Scope::Previous,
-            biscuit_parser::builder::Scope::PublicKey(pk) => {
-                Scope::PublicKey(PublicKey::from_bytes(&pk).expect("invalid public key"))
-            }
+            biscuit_parser::builder::Scope::PublicKey(pk) => Scope::PublicKey(
+                PublicKey::from_bytes(
+                    &pk.key,
+                    match pk.algorithm {
+                        biscuit_parser::builder::Algorithm::Ed25519 => Algorithm::Ed25519,
```
divarvel, line 777: maybe a pair From<> instance for converting between builder algorithm and datalog algorithm would be nicer

biscuit-auth/src/token/builder.rs
```diff
@@ -1611,8 +1651,19 @@ impl From<biscuit_parser::builder::Rule> for Rule {
                     .map(|(k, v)| {
                         (
                             k,
-                            v.map(|bytes| {
-                                PublicKey::from_bytes(&bytes).expect("invalid public key")
+                            v.map(|pk| {
+                                PublicKey::from_bytes(
+                                    &pk.key,
+                                    match pk.algorithm {
+                                        biscuit_parser::builder::Algorithm::Ed25519 => {
```
divarvel, line 1658: same as above

biscuit-parser/src/builder.rs
```diff
@@ -387,7 +392,17 @@ impl ToTokens for Binary {
     }
 }
 
-pub type PublicKey = Vec<u8>;
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct PublicKey {
```
divarvel, line 396: if we want stricter parsing, we could have an enum with proper `PublicKey` values instead of a `Vec<u8>` and an algorithm enum, but that could be done in another PR

biscuit-parser/src/builder.rs
```diff
@@ -157,10 +157,15 @@ impl ToTokens for Scope {
...
             Scope::Previous => quote! { ::biscuit_auth::builder::Scope::Previous},
             Scope::PublicKey(pk) => {
-                let bytes = pk.iter();
-                quote! { ::biscuit_auth::builder::Scope::PublicKey(
-                  ::biscuit_auth::PublicKey::from_bytes(&[#(#bytes),*]).unwrap()
-                )}
+                let bytes = pk.key.iter();
+                match pk.algorithm {
+                    Algorithm::Ed25519 => quote! { ::biscuit_auth::builder::Scope::PublicKey(
+                        ::biscuit_auth::PublicKey::from_bytes(&[#(#bytes),*], ::biscuit_auth::builder::Algorithm::Ed25519).unwrap()
```
divarvel, line 163: the coverage issue could be fixed by adding a relevant test in the macros module

biscuit-auth/src/token/unverified.rs
```diff
@@ -99,7 +101,8 @@ impl UnverifiedBiscuit {
     /// since the public key is integrated into the token, the keypair can be
     /// discarded right after calling this function
     pub fn append(&self, block_builder: BlockBuilder) -> Result<Self, error::Token> {
```
divarvel, line 103: i assume the primary use-case for ECDSA is for the authority block and external signatures, and that we’re ok using ed25519 for attenuation blocks

biscuit-auth/src/token/unverified.rs
```diff
@@ -294,37 +297,44 @@ impl UnverifiedBiscuit {
...
-                ))
+        let algorithm =
+            Algorithm::from_i32(external_signature.public_key.algorithm).ok_or_else(|| {
+                error::Format::DeserializationError(
+                    "deserialization error: invalid external key algorithm".to_string(),
+                )
             })?;
+        let external_key = PublicKey::from_bytes(
+            &external_signature.public_key.key,
+            match algorithm {
```
divarvel, line 325: from instance would be nice here

biscuit-auth/src/token/mod.rs
```diff
@@ -252,6 +249,25 @@ impl Biscuit {
...
+        Self::new_with_key_pair(
+            root_key_id,
+            root,
+            &KeyPair::new_with_rng(builder::Algorithm::Ed25519, rng),
+            symbols,
+            authority,
+        )
+    }
+
+    /// creates a new token, using a provided CSPRNG
```
divarvel, line 264:
```suggestion
    /// creates a new token, using provided keypairs (the root keypair, and the keypair used to sign the next block)
```

biscuit-auth/src/capi.rs
```diff
@@ -310,9 +317,13 @@ pub unsafe extern "C" fn key_pair_new<'a>(
     seed.copy_from_slice(slice);
 
     let mut rng: StdRng = SeedableRng::from_seed(seed);
+    let algorithm = match algorithm {
```
divarvel, line 320: a from instance would be nice

biscuit-auth/Cargo.toml
```diff
@@ -50,6 +50,10 @@ biscuit-parser = { version = "0.1.2", path = "../biscuit-parser" }
 biscuit-quote = { version = "0.2.2", optional = true, path = "../biscuit-quote" }
 chrono = { version = "0.4.26", optional = true, default-features = false, features = ["serde"] }
 serde_json = "1.0.117"
+ecdsa = { version = "0.16.9", features = ["signing", "verifying", "pem", "alloc", "pkcs8", "serde"] }
```
divarvel, line 53: should we feature-gate ecdsa support?

biscuit-auth/src/crypto/mod.rs
Geal: fixed in ae3d779

biscuit-auth/src/token/builder.rs
Geal: fixed in [b0490ae](https://github.com/biscuit-auth/biscuit-rust/pull/108/commits/b0490ae97caa6e442c02b357f1097d7657c8c440)

biscuit-auth/src/token/builder.rs
Geal: done in bdb8c74

biscuit-auth/src/token/builder.rs
Geal: done in https://github.com/biscuit-auth/biscuit-rust/commit/bdb8c7414107ddb16c8cd9b1bd6d4f0f0d0240f5

biscuit-auth/src/token/unverified.rs
Geal: done in https://github.com/biscuit-auth/biscuit-rust/commit/bdb8c7414107ddb16c8cd9b1bd6d4f0f0d0240f5

biscuit-auth/src/capi.rs
Geal: fixing that one later, because capi had a lot of changes in main, that we first need to forward port to v5

biscuit-auth/src/token/unverified.rs
Geal: if people are using ECDSA somewhere, it is likely that they would use it everywhere. Would you prefer that the algorithm stays the same over the entire token?

biscuit-auth/Cargo.toml
Geal: we can, but what would it bring? Reducing the number of dependencies?

biscuit-auth/src/token/unverified.rs
divarvel: i assume the need for using ecdsa is mostly for sensitive keys that need hardware support (eg HSM / smart card, etc). Block keys are not as sensitive as they’re single use, so i guess defaulting to ed25519 is ok here. Also, block keys need the secret key to be accessed directly and stored in the token, so that would not work with a HSM

divarvel (approved):

PR #109, opened 2022-10-21
feat(quote): allow appending to an existing builder

tilpner:
This PR alters the biscuit-quote macros to support *appending* to existing builders, instead of creating new ones. The goal is to allow composition of facts and checks, instead of having to create the entire biscuit in one place.

It is currently possible to sort-of append biscuit builders, by using the `block!` macro and appending that to a `BiscuitBuilder`, but that causes avoidable allocations, and doesn't parse policies. With this PR,

```rust
b.append(block!(
  r#"
    fact(42);
  "#
));
```

becomes

```rust
biscuit!(
  &mut b,
  r#"
    fact(42);
  "#
);
```

- The macros evaluate to values of different types now, depending on if the caller passes a builder. With a builder it returns a `&mut B`, but a `B` without it.
- Distinguishing between clean creation and appending by looking at the first token type feels a bit wrong.
  - I didn't want to have duplicate versions of all three macros for appending.
  - Is there a better syntax?
- The implementation in `ParsedQuery` and `ParsedBiscuitQuery` is duplicated, and should probably be merged into one. I left it as-is for the moment, in case you want to evolve the syntax of them in different directions, but would be happy to adjust the PR to eliminate one of them.

### DIFF
--- a/biscuit-auth/src/token/authorizer.rs
+++ b/biscuit-auth/src/token/authorizer.rs
@@ -260,8 +260,16 @@ impl<'t> Authorizer<'t> {
             */
     }
 
-    pub fn append(&mut self, other: BlockBuilder) {
-        self.authorizer_block_builder.append(other)
+    /// Add the rules, facts, checks, and policies of another `Authorizer`.
+    /// If a token has already been added to `other`, it is not merged into `self`.
+    pub fn merge(&mut self, mut other: Authorizer) {
+        self.merge_block(other.authorizer_block_builder);
+        self.policies.append(&mut other.policies);
+    }
+
+    /// Add the rules, facts, and checks of another `BlockBuilder`.
+    pub fn merge_block(&mut self, other: BlockBuilder) {
+        self.authorizer_block_builder.merge(other)
     }
 
     pub fn add_fact<F: TryInto<Fact>>(&mut self, fact: F) -> Result<(), error::Token>
--- a/biscuit-auth/src/token/authorizer.rs
+++ b/biscuit-auth/src/token/authorizer.rs
@@ -39,7 +39,7 @@ pub struct Authorizer<'t> {
 
 impl<'t> Authorizer<'t> {
     pub(crate) fn from_token(token: &'t Biscuit) -> Result<Self, error::Token> {
-        let mut v = Authorizer::new()?;
+        let mut v = Authorizer::new();
         v.add_token(token)?;
 
         Ok(v)
@@ -54,12 +54,12 @@ impl<'t> Authorizer<'t> {
     /// In the latter case, we can create an empty authorizer, load it
     /// with the facts, rules and checks, and each time a token must be checked,
     /// clone the authorizer and load the token with [`Authorizer::add_token`]
-    pub fn new() -> Result<Self, error::Logic> {
+    pub fn new() -> Self {
         let world = datalog::World::new();
         let symbols = super::default_symbol_table();
         let authorizer_block_builder = BlockBuilder::new();
 
-        Ok(Authorizer {
+        Authorizer {
             authorizer_block_builder,
             world,
             symbols,
@@ -68,7 +68,7 @@ impl<'t> Authorizer<'t> {
             token: None,
             blocks: vec![],
             public_key_to_block_id: HashMap::new(),
-        })
+        }
     }
 
     /// creates an `Authorizer` from a serialized [crate::format::schema::AuthorizerPolicies]
@@ -86,7 +86,7 @@ impl<'t> Authorizer<'t> {
             policies,
         } = crate::format::convert::proto_authorizer_to_authorizer(&data)?;
 
-        let mut authorizer = Self::new()?;
+        let mut authorizer = Self::new();
 
         for fact in facts {
             authorizer
@@ -292,7 +292,7 @@ impl<'t> Authorizer<'t> {
     ///
     /// use biscuit::Authorizer;
     ///
-    /// let mut authorizer = Authorizer::new().unwrap();
+    /// let mut authorizer = Authorizer::new();
     ///
     /// authorizer.add_code(r#"
     ///   resource("/file1.txt");
@@ -1164,14 +1164,14 @@ mod tests {
 
     #[test]
     fn empty_authorizer() {
-        let mut authorizer = Authorizer::new().unwrap();
+        let mut authorizer = Authorizer::new();
         authorizer.add_policy("allow if true").unwrap();
         assert_eq!(authorizer.authorize(), Ok(0));
     }
 
     #[test]
     fn parameter_substitution() {
-        let mut authorizer = Authorizer::new().unwrap();
+        let mut authorizer = Authorizer::new();
         let mut params = HashMap::new();
         params.insert("p1".to_string(), "value".into());
         params.insert("p2".to_string(), 0i64.into());
@@ -1201,7 +1201,7 @@ mod tests {
 
     #[test]
     fn forbid_unbound_parameters() {
-        let mut builder = Authorizer::new().unwrap();
+        let mut builder = Authorizer::new();
 
         let mut fact = Fact::try_from("fact({p1}, {p4})").unwrap();
         fact.set("p1", "hello").unwrap();
@@ -1259,7 +1259,7 @@ mod tests {
 
     #[test]
     fn forbid_unbound_parameters_in_add_code() {
-        let mut builder = Authorizer::new().unwrap();
+        let mut builder = Authorizer::new();
         let mut params = HashMap::new();
         params.insert("p1".to_string(), "hello".into());
         params.insert("p2".to_string(), 1i64.into());
@@ -1355,7 +1355,7 @@ mod tests {
         let res = req.create_block(&external.private(), builder).unwrap();
         let biscuit2 = biscuit1.append_third_party(external.public(), res).unwrap();
 
-        let mut authorizer = Authorizer::new().unwrap();
+        let mut authorizer = Authorizer::new();
         let external2 = KeyPair::new();
 
         let mut scope_params = HashMap::new();
--- a/biscuit-quote/src/lib.rs
+++ b/biscuit-quote/src/lib.rs
@@ -44,7 +44,7 @@ use biscuit_parser::{
     error,
     parser::{parse_block_source, parse_source},
 };
-use proc_macro2::TokenStream;
+use proc_macro2::{Span, TokenStream};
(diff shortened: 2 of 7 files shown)


divarvel:
The original macro design strove for consistency, at the cost of allocating a new builder when appending through a macro (regular fact, rules & checks can still be done). I'm not a fan of (unconstrained) function overloading in general, and macros tend to make things worse in this regard. otoh, unneeded allocations are bad.

I do see the merit of avoiding allocations though. Maybe adding macros for facts, rules & checks directly would allow to use both macros and still append directly without allocating a new block builder?

> It is currently possible to sort-of append biscuit builders, by using the block! macro and appending that to a BiscuitBuilder, but that causes avoidable allocations, and doesn't parse policies.

Policies can't be part of a block (in that case, the authority block), so the parsing part of the `biscuit!` macro is equivalent of the parsing part of the `block!` macro.

tilpner:
> I'm not a fan of (unconstrained) function overloading in general, and macros tend to make things worse in this regard. otoh, unneeded allocations are bad.

If you prefer separate macros (3 -> 6) and have a naming suggestion, I can implement those and try to share as much of the proc macro code as sensible.

> I do see the merit of avoiding allocations though. Maybe adding macros for facts, rules & checks directly would allow to use both macros and still append directly without allocating a new block builder?

I'm not sure what you mean. Could you show an example usage of those hypothetical macros?

> Policies can't be part of a block (in that case, the authority block), so the parsing part of the biscuit! macro is equivalent of the parsing part of the block! macro.

Yes, so since `Authorizer::append` takes a `BlockBuilder`, the pattern of `b.append(block!(...));` can't be used exclusively to iteratively construct authorizers. Unless I'm missing something, the policies either need to be part of the initial `authorizer!` invocation, or added with `Authorizer::add_policy`/`Authorizer::add_code` (both of which lag behind the macros in usability).

divarvel:
Right, i forgot that `Authorizer::append` was taking a `BlockBuilder`, I get what you mean :+1:

What I had in mind for the hypothetical macros was `block_builder.add_fact(fact!(r"fact(true)"));`, but that's not on par with what you've suggested. The current API for individual statements is returning `Result`s, so it's not convenient to use it for things that have guaranteed well-formedness.

All things considered, I think your proposed solution offers the best tradeoffs. The small input and return types inconsistencies can be covered with documentation and comprehensive examples. Separate macros for appending could work as well, but a clear naming would be paramount (since `append` is already used for appending blocks, it would be confusing to re-use it for something else). Maybe `add_to_biscuit!()` (or even `add_to_biscuit_authority!()`) / `add_to_block!()` / `add_to_authorizer!()` would work as well, idk.

As for the way the macro code is currently laid out, it was written in anger, so it could clearly be improved.

Geal:
I'm not a fan of having varying types on `biscuit!`. Could we consider a `BlockBuilder::merge` method instead? That way, multiple block parts can be created independently, some of them could even belong lived, then merged in one block that is appended to the token? I'm fine with `merge` either taking a `&mut self` or returning a new instance.

We should probably have an `AuthorizerBuilder` for consistency. Arguably, `AuthorizerPolicies` was supposed to be that, but right now it's not very usable

tilpner:
> Could we consider a BlockBuilder::merge method instead? 

Are you suggesting a `BlockBuilder::merge` function in addition to separate macros, or instead of them?

> That way, multiple block parts can be created independently, some of them could even belong lived, then merged in one block that is appended to the token?

I'm not so concerned with block reuse as I am with the unnecessary allocations. Consider for example

```rust
let mut b = biscuit!("...");

for fact in facts {
  // add some facts to the biscuit for every `fact`
  // ...
}
```

where facts might be queried from the database, and each loop iteration is supposed to add facts to the biscuit builder.
How would the loop body be implemented? Currently, it would have to `b.append(block!("..."));`, which will create three Vecs (some of them empty and thus cheap) per iteration, only to drain them into the corresponding Vecs of `b`, and then free the Vecs again.

Geal:
Would separate macros for fact, check etc make sense then? I was considering adding them for convenience

divarvel:
it would be nice, but currently there are no infallible methods on builders for already-built datalog elements, so they're a bit less nice than the existing block-level macros (the block-level macros call the `add_` methods and call unwrap themselves)

divarvel:
I think the best path forward is to:

- instead of overloading the existing `biscuit!()`, `block!()` and `authorizer!()` macros, introduce `biscuit_merge!()`, `block_merge!()`, and `authorizer_merge!()` macros
- rename `BlockBuilder::append` to `BlockBuilder::merge` since it's confusing wrt `Biscuit::append` (I can handle it in another PR, it's my bad naming choices, i can fix it myself)
- rename `Authorizer::append` to `Authorizer::merge`, make it take an `Authorizer` and have it return a Result (this way it can error if both authorizers are already carrying a token), and add a `Authorizer::merge_block` taking a `BlockBuilder`). Again, i can handle this part

I think this covers both ergonomic issues and allocation issues. wdyt @tilpner ?

tilpner:
> instead of overloading the existing biscuit!(), block!() and authorizer!() macros, introduce biscuit_merge!(), block_merge!(), and authorizer_merge!() macros

This is now implemented, and the macro implementation is a lot less duplicated. The `authorizer!` and `biscuit!` macros now additionally support the missing/extraneous parameter validation (previously only on `block!`), and parameter validation also accounts for parameters used in policies.

> rename BlockBuilder::append to BlockBuilder::merge since it's confusing wrt Biscuit::append

That's a good idea, I've renamed them.

> rename Authorizer::append to Authorizer::merge, make it take an Authorizer and have it return a Result (this way it can error if both authorizers are already carrying a token), and add a Authorizer::merge_block taking a BlockBuilder).

Done, but I'm not entirely sure I'm merging the correct fields. For now, in addition to merging the block builder, it also copies any policies, but there might be fields that should be merged, which currently aren't.

biscuit-auth/src/token/authorizer.rs
```diff
@@ -260,8 +260,13 @@ impl<'t> Authorizer<'t> {
             */
     }
 
-    pub fn append(&mut self, other: BlockBuilder) {
-        self.authorizer_block_builder.append(other)
+    pub fn merge(&mut self, mut other: Authorizer) {
+        self.merge_block(other.authorizer_block_builder);
+        self.policies.append(&mut other.policies);
```
divarvel, line 265: Thinking about it, a docblock stating that if `other` is already linked to a token, then it'll be discarded should be enough, no need to make `merge` return a `Result`

tilpner: Right! I forgot to ask for clarification about that, as I didn't understand why that would be a problem :/

divarvel:
I really like how it looks like now

biscuit-auth/src/token/authorizer.rs
tilpner: I've added a doc comment to point out that the token is discarded.

tilpner:
There are two things I'm still hoping to address:

- `set_macro_param` (and the subsequent `set_lenient`) are called for every item with every parameters. I don't think this is necessary, and we probably can provide just the relevant parameters
- @meh suggested supporting `let bar = 42; biscuit!("foo({bar}");` like the `println!`/`format!` macros do

tilpner:
Why is biscuit-quote using proc_macro instead of proc_macro2, and would you mind if I changed it to proc_macro2? It's the reason why `quote::__private::*` is needed.

divarvel:
The macro code was my first foray into rust macros, so it might not be up-to-date wrt best practices. Feel free to improve, I'll have a look afterwards to learn a couple things.

Thanks for the improvements!

tilpner:
I've updated both branches to use proc-macro2. This gets rid of the quote::__private shenanigans, and should be free as proc-macro2 was already an indirect dependency.

divarvel:
This looks really good! CI failures are spurious, it's because of the timeout check on token validation combined with poor perf on GHA workers.

tilpner:
I've relaxed the version requirement of proc-macro2 to just "1". I'm still unsure if you'd prefer [these changes](https://github.com/biscuit-auth/biscuit-rust/pull/109#issuecomment-1290543971) (https://github.com/tilpner/biscuit/commit/8426e4b6aac3993154243a9c914edfef1de45364) in a separate PR, or in this one (since it's related, but this one can also stand alone, if you think that makes for easier review).

Geal:
this looks great!

> There are two things I'm still hoping to address:
>
> - set_macro_param (and the subsequent set_lenient) are called for every item with every parameter. I don't think this is necessary, and we probably can provide just the relevant parameters

that can be improved but is not a big issue right now, there's not a huge amount of code generated

> - @meh suggested supporting let bar = 42; biscuit!("foo({bar}"); like the println!/format! macros do

that is a good idea, it can come in another PR

> - parameter value expressions are evaluated multiple times, leading to compile time errors if the parameter is used multiple times and the expression moves a value. this can be fixed by evaluating each parameter once, and then cloning as necessary
itmakes more sense to have the parameter move instead of hiding a clone, that would look more like what function calls taking ownership do

Geal (approved):
this is good to merge now, so we can either integrate it right now or wait for the remaining points

biscuit-quote/src/lib.rs
```diff
@@ -456,83 +456,184 @@ impl Builder {
...
                 .cloned()
                 .collect();
-            let missing_parameters: Vec<String> = self
-                .datalog_parameters
-                .difference(&self.macro_parameters)
-                .cloned()
-                .collect();
             Err(error::LanguageError::Parameters {
-                missing_parameters,
+                missing_parameters: Vec::new(),
```
tilpner, line 468: There is no more fitting error variant for just unused parameters. For a better error message, one could be added.

biscuit-quote/src/lib.rs
```diff
@@ -456,83 +456,184 @@ impl Builder {
...
+    fn fact(fact: &Fact) -> Self {
+        Self {
+            parameters: fact
+                .parameters
+                .iter()
+                .flatten()
+                .map(|(name, _)| name.to_owned())
+                .collect(),
+            start: quote! {
+                let mut __biscuit_auth_item = #fact;
```
tilpner, line 492: These prefixed names are necessary, so that in

```rust
biscuit!("foo({item}, {builder})", item = 42, builder = "foo");
```

the builder or constructed item aren't shadowing user parameters.

Geal: right, that makes sense

Geal:
this looks good, once CI passes I'll merge it, thanks a lot!

PR #110, opened 2022-10-21
feat: support UUIDs as params

tilpner:
Our application uses UUIDs instead of sequential integers as keys, and with the `biscuit-quote` macros it would be an improvement in readability if no manual conversion were necessary:

```rust
biscuit!(
  r#"
    some:id({id});
  "#,
  id = some.id.as_bytes().to_vec(),
  // becomes
  id = some.id,
)
```

Clearly, adding `ToAnyParam` impls for every type that users might want to use is unrealistic. But perhaps Uuids are widespread enough to warrant inclusion.

Is there a way to allow external conversion impls without wrapper types?
 
<details>
  <summary>Technically possible, but really bad idea</summary>

  ```rust
enum AnyParam {
    Bytes(Vec<u8>)
}

trait ToAnyParam {
    fn to_any_param(&self) -> AnyParam;
}

impl ToAnyParam for Vec<u8> {
    fn to_any_param(&self) -> AnyParam { AnyParam::Bytes(self.clone()) }
}

trait SeparateAnyParam {
    fn to_any_param(&self) -> AnyParam;
}

impl SeparateAnyParam for uuid::Uuid {
    fn to_any_param(&self) -> AnyParam { AnyParam::Bytes(self.as_bytes().to_vec()) }
}

fn foo() {
    let b = vec![1, 2, 3].to_any_param();

    // with a macro, we don't *have* to use trait constraints
    // (although we usually should, for error messages)
    let u = uuid::Uuid::new_v4().to_any_param();
}
```
</details>

### DIFF
--- a/biscuit-auth/Cargo.toml
+++ b/biscuit-auth/Cargo.toml
@@ -21,6 +21,7 @@ serde-error = ["serde", "biscuit-parser/serde-error"]
 # used by biscuit-quote to parse datalog at compile-time
 datalog-macro = ["biscuit-quote"]
 docsrs = []
+uuid = ["dep:uuid"]
 
 [dependencies]
 rand_core = "^0.5"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by divarvel** [comment]
I agree that `UUID`s are widespread enough for direct inclusion. For the rest, wrapper types sound like a good solution for that (but that might be the haskeller talking)

**Comment by Geal** [comment]
it would have been nice to have

```rust
impl<T> ToAnyParam for T
where
    T: Into<Vec<u8>>
```

but the compiler has an explicit error for that :smile:

Geal:
Thanks!

PR #111, opened 2022-10-21
add bitwise AND, OR, XOR for ints

nwalsh1995:
Adds support for the following operators on integers:

Bitwise AND: `&`
Bitwise OR: `|`
Bitwise XOR: `^`

Rationale: instead of using a set of actions as strings, we model our permissions as bitflags stored in an integer. To make biscuits more ergonomic it would be helpful if it supported bitwise operations on integers. 

Then we can add checks like 
```
permissions(1011);
check if ($permissions & (1000 | 0001)) > 0
```

Note: I'm not sure if I missed any additional tests, please let me know I'd be happy to add them.


biscuit-auth/src/datalog/expression.rs
```diff
@@ -65,6 +65,9 @@ pub enum Binary {
     Sub,
     Mul,
     Div,
+    BitwiseAnd,
+    BitwiseOr,
+    BitwiseXor,
```
divarvel, line 70: out of consistency with the protobuf and the parser definitions, these could be put at the end of the list

divarvel:
The change itself is a good idea indeed. That would technically be a spec change, and warrant a new block version. Since we're currently working on v4 blocks, we could ship this update along the others (that would require adding a bit of code in the protobuf serialization layer to use the correct block version when these operators are used).

So if you have other ideas / needs for datalog updates, we could discuss it to make sure we can ship everything (that makes sense) in the next update

biscuit-auth/src/datalog/expression.rs
divarvel: :bow:

Geal:
adding those operations make sense. Another test should be added in https://github.com/biscuit-auth/biscuit-rust/blob/master/biscuit-auth/examples/testcases.rs#L1627-L1729 so it can be part of the samples provided with the specification, and as @divarvel we should modify the spec to add this, and it should be good to go!

divarvel:
it would also require a check during unmarshalling that these operators are not used in v3 blocks, and during marshalling that sets the version to 4 if these operators are used:

see https://github.com/biscuit-auth/biscuit-rust/blob/master/biscuit-auth/src/format/convert.rs#L86 and https://github.com/biscuit-auth/biscuit-rust/blob/master/biscuit-auth/src/token/builder.rs#L223

nwalsh1995:
I created some helpers to check schema version conflicts, PTAL!

divarvel:
I've opened a spec PR for the changes https://github.com/biscuit-auth/biscuit/pull/112 (the part about version selection is not written down, as the gist of it is part of the (yet unmerged) 3rd party blocks PR)

Geal:
I modified the version helpers to use boolean flags, which will be easier to handle in the long run. Those helpers really simplify version checks, that's a neat idea!

divarvel:
I think the last issue is wrt precedence. It would be best to be consistent with how it's done in other languages (for instance rust: https://doc.rust-lang.org/reference/expressions.html#expression-precedence)

divarvel:
I have put operators on their own precedence levels https://github.com/nwalsh1995/biscuit-rust/pull/1

Geal:
closing this as it's been integrated in #122. Thanks for the help!

PR #115, opened 2022-10-30
the default rule scope includes the current block

### LINKED ISSUE #114: failing sample execution
the "block_rules" sample is not executed properly:

https://github.com/biscuit-auth/biscuit-rust/actions/runs/3357118911/jobs/5562713872#step:6:83

Geal:
fix #114

### DIFF
--- a/biscuit-auth/src/datalog/origin.rs
+++ b/biscuit-auth/src/datalog/origin.rs
@@ -74,6 +74,7 @@ impl TrustedOrigins {
         if rule_scopes.is_empty() {
             let mut origins = default_origins.clone();
             origins.0.insert(current_block);
+            origins.0.insert(usize::MAX);
             return origins;
         }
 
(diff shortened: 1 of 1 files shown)


divarvel:
the authorizer and the current block are _always_ trusted, not just when there is no scope annotation.

biscuit-auth/src/datalog/origin.rs
```diff
@@ -72,7 +72,9 @@ impl TrustedOrigins {
         public_key_to_block_id: &HashMap<usize, Vec<usize>>,
     ) -> TrustedOrigins {
         if rule_scopes.is_empty() {
-            return default_origins.clone();
+            let mut origins = default_origins.clone();
+            origins.0.insert(current_block);
```
divarvel, line 76:
```suggestion
            origins.0.insert(current_block);
            origins.0.insert(usize::MAX);
```

PR #121, opened 2022-11-22
Fix operator precedences

divarvel:
- [x] Give `&&` and `||` different precedences
- [x] Give `.length()` the same precedence as binary methods
- [x] Give `!` stronger precedence than methods
- [x] Prevent mixing same-precedence non associative operators (`==`, `>=`, `<=`, `<`, `>`)

See https://github.com/biscuit-auth/biscuit/issues/116


biscuit-parser/src/parser.rs
```diff
@@ -515,11 +504,22 @@ fn expr1(i: &str) -> IResult<&str, Expr, Error> {
 }
 
 fn expr2(i: &str) -> IResult<&str, Expr, Error> {
+    let initial_input = i;
     let (i, initial) = expr3(i)?;
 
     let (i, remainder) = many0(tuple((preceded(space0, binary_op_2), expr3)))(i)?;
 
-    Ok((i, fold_exprs(initial, remainder)))
+    if remainder.len() < 2 {
```
Geal, line 512: if we know that remainder either has 0 or 1 element, maybe it should be called onl once instead of with `many0`?

biscuit-parser/src/parser.rs
```diff
@@ -531,26 +531,64 @@ fn expr3(i: &str) -> IResult<&str, Expr, Error> {
...
+    Ok((i, (op, arg)))
+}
+
+fn unary_method(i: &str) -> IResult<&str, builder::Unary, Error> {
+    use builder::Unary;
+    let (i, op) = value(Unary::Length, tag("length"))(i)?;
+
+    let (i, _) = char('(')(i)?;
+    let (i, _) = space0(i)?;
+    let (i, _) = space0(i)?;
```
Geal, line 586: why call `space0` twice?

divarvel: it can only be attributed to human error

biscuit-parser/src/parser.rs
divarvel: yeah, i wanted to be the least disruptive i could, but yeah, being explicit in the parser is probably better

biscuit-parser/src/parser.rs
```diff
@@ -1318,14 +1315,14 @@ mod tests {
...
             super::expr("1 > 2 == 3").map(|(i, o)| (i, o.opcodes())),
-            Err(nom::Err::Failure(crate::parser::Error {
-                input: "1 > 2 == 3",
-                code: nom::error::ErrorKind::Satisfy,
-                message: Some(
-                    "Comparison and equality operators can't be mixed without parentheses"
-                        .to_string()
-                ),
-            }))
+            Ok((
```
Geal, line 1318: I'm a bit uncomfortable with that being an  `Ok`. It should be fine when integrated with other parsers that would bail on "==3" though

Geal (approved):

biscuit-parser/src/parser.rs
divarvel: Yes, chucking an eof() would make the parser fail.

PR #126, opened 2022-12-17
rewrite Authorizer::print_world to generate code

Geal:
Previously, we printed something that looks like a structure dump, but that was not easy to understand. This looks more like a code dump, with the origin indicated in comments

### DIFF
--- a/biscuit-auth/src/datalog/origin.rs
+++ b/biscuit-auth/src/datalog/origin.rs
@@ -1,5 +1,6 @@
 use std::collections::BTreeSet;
 use std::collections::HashMap;
+use std::fmt::Display;
 use std::hash::Hash;
 use std::iter::FromIterator;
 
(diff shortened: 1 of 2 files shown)


biscuit-auth/src/datalog/origin.rs
```diff
@@ -119,3 +143,9 @@ impl<'a> FromIterator<&'a usize> for TrustedOrigins {
         Self(iter.into_iter().cloned().collect())
     }
 }
+
+impl Display for TrustedOrigins {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
```
divarvel, line 151: Is it required by the rest of the changes? It could indicate a confusion between the origin of a rule (the block it was defined in), and its trusted origins. Avoiding this confusion is the point of the separated Origin and TrustedOrigin types.

Geal: not required, that was a mistake, fixed in [e3c3467](https://github.com/biscuit-auth/biscuit-rust/pull/126/commits/e3c3467c7b4b5865f8f293f80415a6e25a4b073c)

divarvel: 👍

PR #135, opened 2023-01-29
handle expression execution failure

Geal:
if an expression fails (example: integer overflow), it should be a hard error instead of returning false, because that boolean might be interpreted at a later stage, like in `deny` policies.

With this change, an expression can return either true, false or an execution error

TODO:
- [ ] add tests
- [ ] define a proper error structure instead of a static string

should the error structure for execution errors indicate which kind of error we got? Should there be some context about which rule or expression triggered it?
With this, type errors (like adding an int and a string) become hard errors and fail the entire rule execution, instead just returning false for the expression evaluation. Are there cases where users relied on relaxed type evaluation?

### DIFF
--- a/biscuit-auth/samples/README.md
+++ b/biscuit-auth/samples/README.md
@@ -1660,5 +1660,5 @@ World {
 }
 ```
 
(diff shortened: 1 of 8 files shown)


divarvel:
fwiw, "blow up on evaluation errors" has been the behaviour in biscuit-haskell from the start and I didn't get any complaints.

It might be useful to relax evaluation rules a bit on type errors in some places (eg have == return false on mismatched types instead of erroring out), though. But apart from cases where an argument can be made for supporting heterogeneous types, i think blowing up is okay

Geal:
Right, that makes sense

PR #139, opened 2023-03-24
biscuit-auth-0.3.0 / biscuit-quote-0.2.0 / biscuit-parser-0.1.0

divarvel:


Geal:
Could you make a version of the PR without the version number changes? Since it's a major release with kind of circular dependencies, the release process will be a bit weird, I'll handle it

divarvel:
I have removed the version bumps.

So, in addition to bumping version numbers in cargo config files, you'll just need to remove the `(Unreleased)` part in the 3 `CHANGELOG.md` files

PR #141, opened 2023-03-24
add `impl From<BTreeSet<Term>> for Term`

divarvel:
This makes it a bit nicer to work with set values.

My initial plan was to have it work with any iterator whose item type is `Into<Term>`, but i can't add a blanket implementation on any iterator.
My second attempt was to add `fn as_term_set<T: Into<Term>, I: Iterator<Item = T>(values: I) -> BTreeSet<Term> {}` function, but turning regular collections into iterators introduces references, and we usually dont have `Into<Term>` instances for those.

If you have any idea for a nice way to turn arbitrary collections into sets of terms, i'm all ears.

### DIFF
--- a/biscuit-auth/src/token/builder.rs
+++ b/biscuit-auth/src/token/builder.rs
@@ -2037,6 +2037,19 @@ impl TryFrom<Term> for SystemTime {
     }
 }
 
+impl From<BTreeSet<Term>> for Term {
+    fn from(value: BTreeSet<Term>) -> Term {
+        set(value)
+    }
+}
+
+#[cfg(feature = "datalog-macro")]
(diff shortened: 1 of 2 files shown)


Geal:
Why not accept `Iterator<Item = Term>`? If I have an iterator of `Into<Term>`, converting them is easy

divarvel:
> Why not accept `Iterator<Item = Term>`? If I have an iterator of `Into<Term>`, converting them is easy

If we already have a `Iterator<Item = Term>` , getting a `BTreeSet<Term>` is just a matter of calling `collect`, so this doesn't buy us much. My goal was to handle the `map(|x| x.into())` so that regular values can be used almost directly (eg using a `Vec<String>` )

PR #149, opened 2023-04-18
fix boolean equality

### LINKED ISSUE #147: Datalog boolean expression == does not evaluate properly
The following datalog code does not produce a bar("d") fact.

```
foo(true);

bar("d") <- foo($b),
$b == true;
```

For info this one does:
```
foo(true);

bar("d") <- foo($b),
$b;
```

Geal:
Fix #147

Equality and inequality were not implemnted for booleans, and there was an invalid catch all for those operations, introduced in b723c633


divarvel:
Can you add a test in the samples suite as well?

I intend to audit all operations tomorrow to see if we've missed something else

divarvel:
I do think the equality catch all made sense though. That's the behavior in many untyped languages, as well as in java for instance

PR #153, opened 2023-04-27
Implement chained method calls

### LINKED ISSUE #146: Parsing chained method calls
The following doesn't parse

```
[1].intersection([1]).contains(1);
```

The following parses:

```
([1].intersection([1])).contains(1);
```

i have noticed the same issue in biscuit haskell, so i suspect it's related to how the parsers are written. i'm not sure yet how hard it would be to fix (to make chained methods left-associative instead of right-associative). In any case, the `Display` implementation on expressions correctly parenthesizes the expression (since it is not aware of precedence rules, it adds parens everywhere, to be sure).

So it's not something someone could copy from generated datalog code, but it's still something that could be expected to work.

Geal:
Fix #146

This is a quick hack, at this point maybe I should look at a proper solution for left recursion in nom


divarvel:
I think it's okay. The construction of the AST itself is left recursive, but the parsing can stay linear as it is.

biscuit-parser/src/parser.rs
```diff
@@ -2167,4 +2171,24 @@ mod tests {
...
+        assert_eq!(
+            super::expr("[1].intersection([1]).contains(1)").map(|(i, o)| (i, o.opcodes())),
+            Ok((
+                "",
+                vec![
+                    Op::Value(set([int(1)].into_iter().collect())),
+                    Op::Value(set([int(1)].into_iter().collect())),
+                    Op::Binary(Binary::Intersection),
+                    Op::Value(int(1)),
+                    Op::Binary(Binary::Contains)
```
divarvel, line 2189: i'd throw in a unary method to make sure parsing both binary and unary methods are correctly handled.

It might be good to add the example to the samples as well.

PR #165, opened 2023-05-25
Add Box trait implementation for RootKeyProvider 

Tipnos:
Addition to enable passing `Box<RootKeyProvider>` to functions with `RootKeyProvider` trait argument. 

Eg: `pub fn from_base64<T, KP>(slice: T, key_provider: KP) -> Result<Self, error::Token>
    where
        T: AsRef<[u8]>,
        KP: RootKeyProvider,`

It is especially useful for middlewares that abstract public keys rotation.

```rust
struct Middleware {
    public_key: Box<dyn RootKeyProvider>
}

impl Middleware{
    fn extract_biscuit(&self, req: ServiceRequest) -> MiddlewareResult<Biscuit> {
        let token = //...

        Biscuit::from_base64(token, self.public_key).map_err(|_e| {
            //...
        })
    }
}
```

### DIFF
--- a/biscuit-auth/src/token/mod.rs
+++ b/biscuit-auth/src/token/mod.rs
@@ -2,6 +2,7 @@
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::fmt::Display;
+use std::rc::Rc;
 
 use self::public_keys::PublicKeys;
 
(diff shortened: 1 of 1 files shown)


Tipnos:
I also added `Rc`and `Arc` implementation to support different middleware implementation.

As an example, actix middleware requires an Rc implementation.

Sorry for late update.

Tipnos:
I'm closing it because other updates might come. 

I'll open a new one when every updates are properly validated.

PR #166, opened 2023-05-26
RootKeyProvider implementation update to enable its use in middleware

Tipnos:
This PR is the continuation of https://github.com/biscuit-auth/biscuit-rust/pull/165 and is related to https://github.com/biscuit-auth/biscuit-actix-middleware/pull/8.

It contains necessary changes to allow use of `RootKeyProvider` in middlewares:
- `Box`, `Rc`, `Arc` implementations (actix uses `Rc`, the rest can be useful for other web frameworks)
- Change `RootKeyProvider` functions argument by `&RootKeyProvider` to  avoid cloning `PublicKey` list foreach call https://github.com/Tipnos/biscuit-rust/commit/fca6c48a722de5941935b11a41bb3cfa40f0efcb

### DIFF
--- a/biscuit-auth/src/format/mod.rs
+++ b/biscuit-auth/src/format/mod.rs
@@ -15,6 +15,7 @@ use crate::crypto::ExternalSignature;
 use crate::datalog::SymbolTable;
 use crate::token::RootKeyProvider;
 use ed25519_dalek::Signer;
+use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::convert::TryInto;
 
@@ -40,13 +41,13 @@ pub struct SerializedBiscuit {
 }
 
 impl SerializedBiscuit {
-    pub fn from_slice<KP>(slice: &[u8], key_provider: &KP) -> Result<Self, error::Format>
+    pub fn from_slice<KP>(slice: &[u8], key_provider: KP) -> Result<Self, error::Format>
     where
-        KP: RootKeyProvider,
+        KP: Borrow<dyn RootKeyProvider>,
     {
         let deser = SerializedBiscuit::deserialize(slice)?;
 
-        let root = key_provider.choose(deser.root_key_id)?;
+        let root = key_provider.borrow().choose(deser.root_key_id)?;
         deser.verify(&root)?;
 
         Ok(deser)
--- a/biscuit-auth/src/format/mod.rs
+++ b/biscuit-auth/src/format/mod.rs
@@ -40,7 +40,7 @@ pub struct SerializedBiscuit {
 }
 
 impl SerializedBiscuit {
-    pub fn from_slice<KP>(slice: &[u8], key_provider: KP) -> Result<Self, error::Format>
+    pub fn from_slice<KP>(slice: &[u8], key_provider: &KP) -> Result<Self, error::Format>
     where
         KP: RootKeyProvider,
     {
--- a/biscuit-auth/src/lib.rs
+++ b/biscuit-auth/src/lib.rs
@@ -59,7 +59,7 @@
 //!   // we want to limit access to `/a/file1.txt` and to read operations
 //!   let token2 = {
 //!     // the token is deserialized, the signature is verified
-//!     let deser = Biscuit::from(&token1,  &root.public())?;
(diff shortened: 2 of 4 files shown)


divarvel:
This is technically a breaking change, since the key provider is now taken by reference. I think it's better now, but that might warrant a major version bump.

biscuit-auth/src/format/mod.rs
```diff
@@ -40,7 +40,7 @@ pub struct SerializedBiscuit {
 }
 
 impl SerializedBiscuit {
-    pub fn from_slice<KP>(slice: &[u8], key_provider: KP) -> Result<Self, error::Format>
+    pub fn from_slice<KP>(slice: &[u8], key_provider: &KP) -> Result<Self, error::Format>
```
Geal, line 43: would this be posible with a `AsRef` to avoid an API breaking change?

Tipnos: No problem! 

Sorry for being picky, would you agree to use `Borrow` instead of `AsRef`? Indeed, `Borrow` has more garantee like `In particular Eq, Ord and Hash must be equivalent for borrowed and owned values: x.borrow() == y.borrow() should give the same result as x == y`

biscuit-auth/src/lib.rs
```diff
@@ -59,7 +59,7 @@
 //!   // we want to limit access to `/a/file1.txt` and to read operations
 //!   let token2 = {
 //!     // the token is deserialized, the signature is verified
-//!     let deser = Biscuit::from(&token1,  root.public())?;
+//!     let deser = Biscuit::from(&token1, root.public())?;
```
divarvel, line 62: it's good to know it works well in both cases, but maybe we should harmonize the examples?

imo passing it by reference should be the default in examples. If we want to make sure it works both ways (and we should!), ensuring it in tests would be good and would not affect documentation.

Tipnos: I put both to show all possibilities as you were concerned by API breaking changes. 

But IMO I find reference only examples better!

I'm waiting for @Geal approval to make final modifications.

Geal (approved):

biscuit-auth/src/lib.rs
Geal: references should work better for the examples yes

Tipnos:
Bad news!

It looks like we can't allow passing `&RootKeyProvider` as function argument without introducing breaking change.

As we saw earlier we have to introduce `Borrow` trait to be able to pass by value and reference. 

But thanks to a nice @divarvel final review catch, closure impl was no more tested. Indeed, I removed the test case by misunderstanding of its use.

To pass the test the following implementation is required:
```rust 
impl<F: Fn(Option<u32>) -> Result<crypto::PublicKey, error::Format>> Borrow<dyn RootKeyProvider>
    for F
{
    fn borrow(&self) -> &(dyn RootKeyProvider + 'static) {
        self
    }
} 
```

[But rust does not allow it](https://rust-lang.github.io/rfcs/2451-re-rebalancing-coherence.html).

An other alternative could be:
```rust
impl Borrow<dyn RootKeyProvider>
    for dyn Fn(Option<u32>) -> Result<crypto::PublicKey, error::Format>
{
    fn borrow(&self) -> &(dyn RootKeyProvider + 'static) {
        self
    }
} 
```

But cast between traits is not allowed.

Final alternative:
```rust
impl Borrow<dyn RootKeyProvider> for fn(Option<u32>) -> Result<crypto::PublicKey, error::Format> {
    fn borrow(&self) -> &(dyn RootKeyProvider + 'static) {
        self
    }
}
```

This implementation compiles but as the closure has an argument (capture its env) it can't be cast as a function pointer.

After discussion with @divarvel, we think the best solution is to put the subject aside until next major release. 

Concerning https://github.com/biscuit-auth/biscuit-actix-middleware/pull/8 we'll clone `Vec<PublicKey>` on every incoming calls until next major release.

PS: if performance issues about cloning occure before next major update, we still have the possibility to temporarily add a new function that take a reference.

Tipnos:
See https://github.com/biscuit-auth/biscuit-rust/pull/168 for final merged implementation.

PR #173, opened 2023-06-28
Support for biscuit web keys

divarvel:
See https://github.com/biscuit-auth/biscuit/blob/master/biscuit-web-key/SPECIFICATIONS.md

This adds `Serialize` and `Deserialize` instances for `BiscuitWebKey`, which groups a `PublicKey`, a mandatory `u32` `key_id` and an optional `issuer` `String`.

Enabling the `bwk` feature adds `serde` and `chrono` as dependencies

### DIFF
--- a/biscuit-auth/src/bwk.rs
+++ b/biscuit-auth/src/bwk.rs
@@ -1,0 +1,209 @@
+use std::convert::TryFrom;
+
+use chrono::{DateTime, FixedOffset};
+use serde::{Deserialize, Serialize};
+
+use crate::{error, PublicKey};
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+#[serde(into = "BiscuitWebKeyRepr")]
+#[serde(try_from = "BiscuitWebKeyRepr")]
+pub struct BiscuitWebKey {
+    pub public_key: PublicKey,
+    pub key_id: u32,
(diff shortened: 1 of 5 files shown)


biscuit-auth/src/lib.rs
```diff
@@ -241,6 +241,11 @@ mod capi;
 #[cfg(cargo_c)]
 pub use capi::*;
 
+#[cfg(bwk)]
+mod bwk;
+#[cfg(bwk)]
+pub use bwk::*;
+
```
Tipnos, line 248:
```suggestion
#[cfg(bwk)]
{
    mod bwk;
    pub use bwk::*;
}

```

biscuit-auth/src/bwk.rs
```diff
@@ -0,0 +1,209 @@
...
+            public_key,
+            key_id: value.key_id,
+            issuer: value.issuer,
+            expires_at: value.expires_at,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
romgus, line 62: Why are the tests in the same file? Is it really convenient mixing test/prod code?

biscuit-auth/src/lib.rs
divarvel: i have reverted it, it is not allowed.

PR #179, opened 2023-08-03
feat: detect free variables in rule expressions

divarvel:
Free variables are already forbidden in rule heads, but not in rule expressions.

# Open questions

The reported error span was the rule head when we were looking for free variables in the rule head. Now free variables can be found in the whole expression, so for now the whole expression is reported as the error span, which is not perfect.
The ideal solution would be to have multiple errors, one per free variable instance, with the span being only the variable itself. That may be hard to achieve, so maybe just being able to keep the expression span (or the rule head span) would be good, but that's still quite a bit of work.

I noticed a fair amount of duplication between the `builder` module in `biscuit-parser` and the one in `biscuit-auth`. For instance, `validate_variables` from biscuit-auth is not used anywhere, only the one from `biscuit-parser`. Could the `biscuit-auth` `builder` module be significantly reduced to use types and functions from `biscuit-parser`?

# ToDo

- checks
- policies
- queries

### DIFF
--- a/biscuit-parser/src/builder.rs
+++ b/biscuit-parser/src/builder.rs
@@ -1,6 +1,6 @@
 //! helper functions and structure to create tokens and blocks
 use std::{
-    collections::{BTreeSet, HashMap},
+    collections::{BTreeSet, HashMap, HashSet},
     time::{SystemTime, UNIX_EPOCH},
 };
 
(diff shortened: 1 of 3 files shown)


Geal (approved):

divarvel:
Maybe we can keep the biscuit-auth types as newtypes if we want to avoid
orphan instances. (It depends on where the traits are defined ofc).

Le lun. 7 août 2023, 21:04, Geoffroy Couprie ***@***.***> a
écrit :

divarvel:
Are you ok with the error span being on the whole rule instead of just the head for head variables? Also for checks & policies, it will require a bit of work to collect scope errors in all rule bodies without aborting on the first rule with free variables.

biscuit-parser/src/builder.rs
```diff
@@ -328,33 +328,38 @@ impl Rule {
     }
 
     pub fn validate_variables(&self) -> Result<(), String> {
```
divarvel, line 330: this would need to return a 3-valued enum (Head, Body, Both) to signal where the free variables were found, in order to only return the relevant span in the parsing error.

PR #181, opened 2023-08-16
add a key argument to test cases

Geal:
this will help in the migration to the new version of ed25519-dalek and rand. Test cases use a RNG with a static seed, but generation changed


divarvel:
wouldn't it be better to have the verification mode read the key from the samples file? That's what other implementations do, since they can't rely on rng consistency with rust, and read everything else from the samples file.

Geal:
I can add it, but it would only work with the json output, not the markdown output

Geal:
Not used by automated tooling, but very useful when debugging the crypto algorithms, as I'm doing right now 😄

Geal:
working from json requires a larger refactoring of the samples, so I'll do that as follow up, and merge this for now, as it will unblock the ed25519-dalek upgrade

PR #182, opened 2023-08-22
make the third party tests more deterministic

Geal:
follow up on https://github.com/biscuit-auth/biscuit-rust/pull/181 to help with the dalek upgrade


Geal:
I think at this point it might just make more sense to have a static root key in the samples instead of passing it as argument

divarvel (approved):

PR #188, opened 2023-10-11
wip: short-circuiting boolean operators

divarvel:
Even though the semantics of the stack machine are eager, what we actually care about is && and || being non-strict (rather than lazy). The difference being non-strict and lazy is not observable from the outside since datalog cannot perform side-effects.

This PR is only intended as a proof of concept, since this update is a breaking change. It would have to be bundled in a token block version upgrade.

A note: unbound variables still abort evaluation. This can be debated. My point of view is that variable bindings are static as far as the expression itself is concerned, while actual evaluation failures usually depend on the actual values bound to variables.

I'm not sure why the samples test pass though, running them locally gives a different output for sample 27, where strict evaluation of booleans is tested.


divarvel:
I think i have found a way to have closures in the stack machine. Laziness is trivially implemented through closures, so this PR might be closed soon.

divarvel:
superseded by #202

PR #189, opened 2023-10-17
Add UnverifiedBiscuit::verify

divarvel:
It takes a `RootKeyProvider` (instead of a closure like `check_signature`).

It avoids introducing a breaking change by keeping `check_signature` as is (but deprecates it). `verify` is more consistent with the terminology anyway.


BastienVigneron:
lgfm

BastienVigneron (approved):
lgfm

Geal (approved):

PR #191, opened 2023-10-26
Better display for `errors::Logic`

divarvel:
```
authorization failed: an allow policy matched (policy index: 0), and the following checks failed: [Authorizer(FailedAuthorizerCheck { check_id: 0, rule: "check if right(\"efgh\")" }), Block(FailedBlockCheck { block_id: 1, check_id: 0, rule: "check if operation(\"read\")" })]
```

### DIFF
--- a/biscuit-auth/src/error.rs
+++ b/biscuit-auth/src/error.rs
@@ -1,7 +1,10 @@
 //! error types
 //!
 
-use std::convert::{From, Infallible};
+use std::{
+    convert::{From, Infallible},
+    fmt::Display,
+};
 use thiserror::Error;
 
 /// the global error type for Biscuit
@@ -16,7 +19,7 @@ pub enum Token {
     AppendOnSealed,
     #[error("tried to seal an already sealed token")]
     AlreadySealed,
-    #[error("authorization failed")]
+    #[error("authorization failed: {0}")]
     FailedLogic(Logic),
     #[error("error generating Datalog: {0}")]
     Language(biscuit_parser::error::LanguageError),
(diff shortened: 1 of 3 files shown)


biscuit-auth/src/error.rs
```diff
@@ -168,9 +171,9 @@ pub enum Signature {
 #[derive(Error, Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "serde-error", derive(serde::Serialize, serde::Deserialize))]
 pub enum Logic {
-    #[error("a rule provided by a block is generating facts with the authority or ambient tag, or has head variables not used in its body")]
+    #[error("a rule provided by a block is producing a fact with unbound variables")]
```
divarvel, line 174: i wanted to completely remove it, but it's still used. the strange thing is that it explicitly mentions block rules. Where is the error for authorizer rules with unbound variables in facts?

divarvel: after a bit of digging:

- unbound variables are checked by the parser when constructing rules, so all rules constructed by the parser will be checked, regardless of their origin, and the error will be a parser error, not a `Logic` error
- this error can indeed only happen when a block is added to the authorizer, so adding a broken rule to the authorizer will not be caught here (rules constructed by the parser will be ok, it can only happen for manually constructed rules)
- rules with unbound variables in head will not generate any fact (but not crash)

I think the ergonomics of this could be improved:

- by being clearer on invariants that `Rule` and friends should guarantee, no matter how they were constructed (this would be a bit of work because i think that we should do it in other places as well)
- for now, by calling `validate_variables` when adding rules to the authorizer, not only rules blocks

Geal: That makes sense, we should avoid those kinds of footguns

PR #197, opened 2023-11-12
make sure that we actually need to run the datalog engine

Geal:
the authorizer needs the datalog engine to have executed before performing certain tasks, but we might have already done it precedently, without having added facts or rules.
This adds a boolean tracking if we need to run: if we added facts or rules, it is set to true, if we have just run the datalog engine, it is set to false

### DIFF
--- a/biscuit-auth/src/token/authorizer.rs
+++ b/biscuit-auth/src/token/authorizer.rs
@@ -40,6 +40,8 @@ pub struct Authorizer {
     public_key_to_block_id: HashMap<usize, Vec<usize>>,
     limits: AuthorizerLimits,
     execution_time: Duration,
+    /// indicates if new data was added and the rules must be executed again
+    dirty: bool,
 }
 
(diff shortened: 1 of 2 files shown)


biscuit-auth/src/token/authorizer.rs
```diff
@@ -218,33 +222,38 @@ impl Authorizer {
     /// Add the rules, facts, checks, and policies of another `Authorizer`.
     /// If a token has already been added to `other`, it is not merged into `self`.
     pub fn merge(&mut self, mut other: Authorizer) {
+        self.dirty = true;
```
divarvel, line 225: this one is redundant, but better safe than sorry, I guess

divarvel (commented):
Looks good, i'd be more comfortable with tests exercising it though

Geal:
superseded by https://github.com/biscuit-auth/biscuit-rust/pull/250

PR #199, opened 2023-12-28
Reject if

divarvel:
See https://github.com/biscuit-auth/biscuit/issues/141

This adds `reject if` support to the parser as well as the serialization layer, raising the block version to 5 when used.

This PR does not introduce an equivalent to `reject if all` because i'm not sure what it would look like and whether it would be useful.


biscuit-auth/src/datalog/mod.rs
```diff
@@ -625,7 +626,9 @@ impl World {
 
             index += 1;
             if index == limits.max_iterations {
-                break Err(Execution::RunLimit( crate::error::RunLimit::TooManyIterations));
+                break Err(Execution::RunLimit(
```
BastienVigneron, line 629: formatting issue ?

biscuit-auth/src/datalog/mod.rs
```diff
@@ -887,7 +896,16 @@ pub fn get_schema_version(
             .iter()
             .any(|c: &Check| c.queries.iter().any(|q| !q.scopes.is_empty()));
 
-    let contains_check_all = checks.iter().any(|c: &Check| c.kind == CheckKind::All);
+    let mut contains_check_all = false;
+    let mut contains_reject_if = false;
+
+    for c in checks.iter() {
```
BastienVigneron, line 902: Couldn't be simplified with a `contains` ?

biscuit-auth/src/datalog/mod.rs
```diff
@@ -986,19 +1005,21 @@ mod tests {
...
+                    &[pred(
+                        parent,
+                        &[var(&mut syms, "parent"), var(&mut syms, "child")],
+                    )],
+                ),
+                0,
+                &[0].iter().collect(),
+                &syms,
+            )
+            .unwrap();
```
BastienVigneron, line 1022: That's why I prefer to move tests in separate files :)
I've clicked 5 times on "Expand all" button to understand we are in a test fn.

biscuit-auth/src/format/convert.rs
```diff
@@ -71,10 +71,19 @@ pub fn proto_block_to_token_block(
...
+    if version < MAX_SCHEMA_VERSION {
+        for c in input.checks_v2.iter() {
+            if version == MIN_SCHEMA_VERSION && c.kind.is_some() {
+                return Err(error::Format::DeserializationError(
+                    "deserialization error: v3 blocks must not contain a check kind".to_string(),
+                ));
+            } else if c.kind == Some(schema::check_v2::Kind::Reject as i32) {
+                return Err(error::Format::DeserializationError(
+                    "deserialization error: v4 blocks must not contain reject if checks"
+                        .to_string(),
```
BastienVigneron, line 83: `to_owned()` ;-)

biscuit-auth/src/datalog/mod.rs
```diff
@@ -214,8 +214,10 @@ impl Rule {
                     Ok(Term::Bool(false)) => {
                         //println!("expr returned {:?}", res);
                         return Ok(false);
-                    },
-                    Ok(_) => return Err(error::Execution::Expression(error::Expression::InvalidType)),
+                    }
+                    Ok(_) => {
```
BastienVigneron, line 218: formatting issue ?

biscuit-auth/src/datalog/mod.rs
divarvel: yeah, there is a commit fixing formatting errors

biscuit-auth/src/datalog/mod.rs
divarvel: we are setting two separate flags, so using `contains` (or `any` as before) would require traversing the list twice.

biscuit-auth/src/datalog/mod.rs
divarvel: yeah, tests could be moved to a separate file at some point. not the topic of this pr tho

biscuit-auth/src/format/convert.rs
divarvel: we're using `to_string()` in the other blocks, so I kept it for consistency.

biscuit-auth/src/token/third_party.rs
```diff
@@ -124,7 +126,7 @@ impl ThirdPartyRequest {
         let mut symbols = SymbolTable::new();
         symbols.public_keys = self.public_keys.clone();
         let mut block = block_builder.build(symbols);
-        block.version = super::MAX_SCHEMA_VERSION;
+        block.version = max(4, block.version);
```
Geal, line 129: we should probably have a list of constants somewhere like `SCHEMA_VERSION_3RD_PARTY = 4` to track those checks more easily

Geal: I can take care of that after merging that one

divarvel: I had the exact same thought, yep

Geal: added in https://github.com/biscuit-auth/biscuit-rust/pull/217/commits/c62aeabda3b0fe98d12630fbe40b327b2b7c0917

Geal (approved):

PR #202, opened 2023-12-29
Support for closures

divarvel:
The goal of this branch is to explore how to add support for closures in the stack machine

- closures are represented as a list of ops and a list of parameter names
- closure parameters can be used like variables
- variable names must be unique, so shadowing is forbidden
- boolean operators use laziness and push the rhs closure on the ops list only if needed (without recursion)
- `.any()` and `.all()` use recursion to compute results

- [x] closure support in `datalog::expression::Op`
- [x] closure support in `format::schema`
- [x] closure support in `parser`
- [x] closures for lazy boolean operators
- [x] closures for `.any()` and `.all()`
- [x] prevent shadowing in closure arguments
  - [x] evaluation
  - ~parsing & deserialization depends on #179~ will be done in a dedicated PR
- [x] print `And` and `Or` with `&&!` and `||!`, without parsing them.
- [x] v5 block detection depends on #220


biscuit-auth/src/datalog/expression.rs
```diff
@@ -238,52 +298,98 @@ impl Binary {
...
-                    _ => {
-                        //println!("expected two values on the stack");
+                    e => {
+                        println!(
+                            "while evaluating {}",
+                            self.print(&SymbolTable::new()).unwrap()
+                        );
+                        println!("while evaluating {opop:?}");
+                        println!("with context {values:?}");
+                        println!("expected two values on the stack, got: {e:?}");
```
divarvel, line 372: todo remove

Geal (approved):
great job!

PR #204, opened 2024-02-03
Expose PEM/DER decoder functions in KeyPair

baranyildirim:
To make it easier to loader KeyPairs, expose PEM and DER loaders from the `ed25519_dalek` crate.
PEM and DER are common formats for a key pair, so this should make loading a KeyPair easier.


divarvel:
Thanks! idk how @Geal feels about it, but I think it should be feature-gated. Apart from that, LGTM

baranyildirim:
> Thanks! idk how @Geal feels about it, but I think it should be feature-gated. Apart from that, LGTM

Added a feature gate called "pem" to match the feature gate in the underlying crate.

Geal (approved):

PR #207, opened 2024-02-13
biscuit-auth 4.1.0

Geal:


divarvel (commented):
Looks good, could you update the changelog as well?

I assume Cargo.lock is not versionned

Geal:
yes, it's not versioned, as it is just a set of libraries

divarvel (approved):

divarvel:
> yes, it's not versioned, as it is just a set of libraries

Recommendations for cargo.lock versioning have changed recently-ish, and now putting it under version control is advised (but not required)

Geal:
ah, I'll look into that then

PR #208, opened 2024-02-13
remove pkcs8 file loading functions

Geal:
We made a mistake when merging https://github.com/biscuit-auth/biscuit-rust/pull/204, that feature can't compile right now, because it needs the `std` feature active on the e25519 crate. That features requires `std::fs`, which may not work easily in other target platforms like web assembly.
It is reasonably easy to load a file manually in rust and pass a buffer, so I propose we remove those functions entirely.

cc @baranyildirim

### DIFF
--- a/biscuit-auth/src/crypto/mod.rs
+++ b/biscuit-auth/src/crypto/mod.rs
@@ -16,8 +16,6 @@ use ed25519_dalek::*;
 
 use nom::Finish;
 use rand_core::{CryptoRng, RngCore};
-#[cfg(feature = "pem")]
-use std::path::Path;
 use std::{convert::TryInto, fmt::Display, hash::Hash, ops::Drop, str::FromStr};
 use zeroize::Zeroize;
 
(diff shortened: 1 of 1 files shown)


divarvel (approved):
Agreed

baranyildirim:
Hey folks, does the gating behind the feature not work here?

Geal:
The gating works, but when activating the pem feature, the library does not compile

baranyildirim:
> The gating works, but when activating the pem feature, the library does not compile

Oh, should we fix forward instead? It sounds like we need to make the pem feature enable the other features that would enable std::fs.

I think pkcs8 is a pretty common format, but I can make my own workaround. Feel free to revert.

Geal:
What I'm proposing in this format keeps the pem and der loading from buffers, but not the functions to load from file, because to activate files I would then have to add a second feature gate to the router, since file loading would not make sense in other platforms like wasm.
If you have the functions to load from a buffer, loading from a file is not too far, that's a reasonable tradeoff

baranyildirim:
Sounds good. I think a pkcs8 PEM file is quite common (easier to use interface), but I don’t mind the change. Sorry for breaking the build.

PR #210, opened 2024-02-14
fix capi build

Geal:


divarvel (approved):

Geal:
this will require some changes to the C API because it cannot propose the same features using biscuit-auth's public API. But also, that API does not really match what we are doing right now, with the focus on `Authorizer`, so that API needs a rewrite

ptondereau:
@Geal still need work to-do?

Geal:
we got some other fixes in the meantime

PR #212, opened 2024-03-10
Add pem/der loaders to PublicKey

baranyildirim:
Similar to #204 
It seems like the previous feature gate does not work correctly - cargo does not detect the `pem` feature unless it is also available in the biscuit-parser crate. 

I assume something similar happened with the `datalog-macro`, so I've added an empty `pem` feature to biscuit-parser.


Geal:
could you tell me more about what is not working properly? which error were you seeing?

Geal:
@baranyildirim the work from this PR was merged as part of https://github.com/biscuit-auth/biscuit-rust/pull/265, so I'm closing this. Thank you for your help!

Issue #211, opened 2024-03-10
Proposal: Use padding free Base64 encoding/decoding


baranyildirim:
Hey folks,

I think I've found an issue with how base64 encoding/decoding works with Biscuit tokens.
Currently, the encode/decode config used for Base64 operations is invoked as follows:
- `base64::encode_config(slice, base64::URL_SAFE)` in  biscuit-auth/src/token
- `base64::decode_config(slice, base64::URL_SAFE)` in  biscuit-auth/src/token

The problem here is that the default encode/decode configs use padding. The padding character for base64 is the `=` character, which is not url-safe.

This means that any token that requires padding cannot be correctly used in contexts with URLs.

For example, take the token on the biscuitsec front page (== at the end):

```
En0KEwoEMTIzNBgDIgkKBwgKEgMYgAgSJAgAEiAs2CFWr5WyHHWEiMhTXxVNw4gP7PlADPaGfr_AQk9WohpA6LZTjFfFhcFQrMsp2O7bOI9BOzP-jIE5PGhha62HDfX4t5FLQivX5rUhH5iTv2c-rd0kDSazrww4cD1UCeytDSIiCiCfMgpVPOuqq371l1wHVhCXoIscKW-wrwiKN80vR_Rfzg==
```

When used with URL encoding, this token becomes:
```
En0KEwoEMTIzNBgDIgkKBwgKEgMYgAgSJAgAEiAs2CFWr5WyHHWEiMhTXxVNw4gP7PlADPaGfr_AQk9WohpA6LZTjFfFhcFQrMsp2O7bOI9BOzP-jIE5PGhha62HDfX4t5FLQivX5rUhH5iTv2c-rd0kDSazrww4cD1UCeytDSIiCiCfMgpVPOuqq371l1wHVhCXoIscKW-wrwiKN80vR_Rfzg%3D%3D
```

Now, this token can no longer be decoded using the biscuit-rust library. For example:
```rust
use biscuit_auth::Biscuit;
[...]

PR #213, opened 2024-03-10
Add pad option to base64 encode/decode functions

baranyildirim:
See #211

### DIFF
--- a/biscuit-auth/src/token/mod.rs
+++ b/biscuit-auth/src/token/mod.rs
@@ -99,12 +99,12 @@ impl Biscuit {
     }
 
     /// deserializes a token and validates the signature using the root public key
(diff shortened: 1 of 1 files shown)


divarvel:
Given that other implementations depend on the padding, we should not modify default behaviour, and since the lib allows to manually manage base64 encoding, I will close this PR.

baranyildirim:
> accepting padded and non padded in deserialization, but only producing non padded base64 should be fine. 

I don't understand how this is breaking compatibility, but I'll just maintain a separate fork. Thanks.

divarvel:
since you can use `Biscuit::from()` with a byte slice directly, creating a helper function with the base64 settings you prefer would be way simpler than maintaining a fork imo

PR #214, opened 2024-03-24
Implement query_exactly_one

### LINKED ISSUE #187: `query_exactly_one()`
In many cases, we want the query to return exactly one result. A canonical example would be querying a user id from a token.

The current API makes it easier to discard extra results than to throw an error in the case of extra results.

I think it would be a good addition to have a query function that only returns a result when there exactly one match

baranyildirim:
Implement query exactly one to make it easier to work with single-fact queries.
Added relevant doctest and unittests. This should close #187 
As a side note, a lot of errors in Token::Error don't expose the underlying error (e.g. RunLimit) in their error message. Is this intentional?

### DIFF
--- a/biscuit-auth/src/error.rs
+++ b/biscuit-auth/src/error.rs
@@ -260,6 +260,8 @@ pub enum RunLimit {
     TooManyIterations,
     #[error("spent too much time verifying")]
     Timeout,
+    #[error("Unexpected query results, expected {0} got {1}")]
+    UnexpectedQueryResult(usize, usize),
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


divarvel:
LGTM.

tbf the whole Token::Error hierarchy would need a refresh

divarvel:
The capi build failure is being worked on, I think

divarvel:
If that’s ok with you, I will rebase the PR on top of `main`

divarvel:
Superseded by #260 

thanks for opening this PR and sorry for the time it stayed open, work on the `v5` branch imposed a kind of freeze on `main`

PR #219, opened 2024-05-23
add new Array and Map types

Geal:
this tries to enforce that all array elements are of the same type, but this is not very strict at the moment, it does not look at lower levels of composite types.

**breaking changes**:
- in the Datalog language, sets will now be delimited by '{' and '}' instead of '[' and ]'. Arrays are now delimited by '[' and ']'
- parameter names must now start with a letter

Todo:
- [x] array serialization
- [x] array parsing
- [x] existing operations for array: length, equal, not equal, contains, prefix, suffix
- [x] map serialization
- [x] map parsing
- [x] existing operations for map: length, equal, not equal, contains, prefix, suffix
- [x] new operations for array: get (with integer index), closures
- [x] new operations for map: get (with integer or string index), closures


biscuit-parser/src/parser.rs
```diff
@@ -728,6 +732,18 @@ fn name(i: &str) -> IResult<&str, &str, Error> {
     reduce(take_while1(is_name_char), " ,:(\n;")(i)
 }
 
+fn parameter_name(i: &str) -> IResult<&str, &str, Error> {
```
divarvel, line 735: this introduces a regression: 1-letter parameter names are not allowed any more

Geal: fixed in 6d3b426

PR #226, opened 2024-06-05
Apply some clippy lint where it makes sense

ptondereau:
No big deals here, and just applying some clippy rules that make sense.
Please don't hesitate to close if it's not relevant


Geal (approved):

Geal:
thanks!

PR #228, opened 2024-10-02
datalog foreign function interface prototype

divarvel:
This allows using external functions in datalog.

This makes it easy to provide custom logic without extending the spec for every use-case, at the expense of portability: behaviour is no longer guaranteed to be consistent cross languages, and some languages won’t be able to support it at all (for instance JS as of now).


biscuit-auth/src/datalog/mod.rs
```diff
@@ -737,6 +743,8 @@ pub struct RunLimits {
     pub max_iterations: u64,
     /// maximum execution time
     pub max_time: Duration,
+
+    pub extern_funcs: HashMap<String, ExternFunc>,
```
Geal, line 747: I am not sure putting the functions in `RunLimits` is the right way. I'd prefer if it was stored at the `World` level, so we can make sure they are always applied

biscuit-auth/src/token/builder.rs
```diff
@@ -419,6 +422,50 @@ pub trait Convert<T>: Sized {
     }
 }
 
+/// Builder for a unary operation
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Unary {
```
Geal, line 427: good call, the datalog version should not be used in builders

Geal (approved):

Geal:
that will be a great tool for future experimentation

PR #229, opened 2024-10-02
fix: recursively collect and apply parameters in closures

divarvel:
Datalog parameters are managed lazily: upon building a rule, referenced parameter are collected in a map. Setting them only updates the map, and the actual AST is only updated as needed (such as in Display impls, or when converting to datalog data structures).

Expressions used to be linear, but closures introduced the possibility of recursive expressions, so the code inside closure bodies was ignored both when collecting parameters, and when applying them.

This caused an issue because some parts of the code assume that unbound parameters are handled beforehand, and panic upon encountering them.

### DIFF
--- a/biscuit-auth/src/token/builder.rs
+++ b/biscuit-auth/src/token/builder.rs
@@ -913,6 +913,43 @@ pub enum Op {
     Closure(Vec<String>, Vec<Op>),
 }
 
+impl Op {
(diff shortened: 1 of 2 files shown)


divarvel:
capi support will be extracted to a dedicated crate, so we can ignore the failing build step.

PR #231, opened 2024-10-21
Major minor datalog versions

divarvel:
fix version checks; introduce x.y versioning for datalog

While the incrementing integer for version numbers is fine for the protobuf encoding, it has been a headache when communicating with other people.

A `major.minor` scheme seems better for understanding, and helps communicate the scope of changes a bit better.

There was a bug in version checks that relied on `BISCUIT_MAX_VERSION` to check the version for specific features. This does not play well with version upgrades.

So this commit fixes the following issues:

- use explicit 3.x version number for datalog features
- use explicit version numbers for feature checks, instead of relying on the max version
- bump the min version for third-party blocks from 3.1 to 3.2
- bump the min version for features of the v5 branch to 3.3

It also improves a bit on some error messages that were a bit cryptic, by clearly specifying the version in which specific features were implemented.


biscuit-auth/samples/samples.json
```diff
@@ -2132,7 +2132,7 @@
           },
           "authorizer_code": "allow if true;\n",
           "revocation_ids": [
-            "365092619226161cf3973343f02c829fe05ab2b0d01f09555272348c9fcce041846be6159badd643aee108c9ce735ca8d12a009979c46b6e2c46e7999824c008"
+            "31f01031b8d9cc312c346674625a2218b19a126e931318dc02d8d53aec2775faecd3f848f60c1446573c37ea0664fefc768f35ba5ce4365b83ff06f0c0409b0e"
```
divarvel, line 2135: this changes because the test uses the heterogeneous not equals, which is now a v3.3+ operator

Geal (approved):

PR #232, opened 2024-10-21
Replace cfg(cargo_c) by cfg(feature="capi") + upgrade cargo-c in the CI

ptondereau:
- Replace cfg(cargo_c) by cfg(feature="capi")
- Upgrade cargo-c

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,76 +2,74 @@ name: Rust
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main, v5 ]
+    branches: [main, v5]
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
   build:
-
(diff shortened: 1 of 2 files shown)


ptondereau:
@Geal it seems that running `cargo ctest` makes [all tests blinds](https://github.com/biscuit-auth/biscuit-rust/actions/runs/11448454611/job/31852977480?pr=232#step:5:302) about the crate `biscuit_auth` and not for `biscuit_quote`...

Running with `-Z macros-backtrace` doesn't help too.

Unfortunately, I've run out of ways to move forward :'(

Geal:
That's the same issue I encountered when I tried to update it. At this point I think the next step would be to try and extract it in a separate crate

ptondereau:
> That's the same issue I encountered when I tried to update it. At this point I think the next step would be to try and extract it in a separate crate

We can merge this one as it is for the moment. I'll move capi to its own crate in the next PR.
WDYT?

PR #234, opened 2024-10-22
refactor(capi): Move CAPI generation to its own crate

ptondereau:
Extracting capi generation into its own crate

Also, I've updated the release workflow to support multiple OS and arch (Linux and MacOS)


biscuit-auth/src/token/mod.rs
```diff
@@ -73,9 +73,9 @@ pub fn default_symbol_table() -> SymbolTable {
 #[derive(Clone, Debug)]
 pub struct Biscuit {
     pub(crate) root_key_id: Option<u32>,
-    pub(crate) authority: schema::Block,
-    pub(crate) blocks: Vec<schema::Block>,
-    pub(crate) symbols: SymbolTable,
+    pub authority: schema::Block,
```
Geal, line 76: not really comfortable with exposing this, it should not be necessary. We'll have to adapt the C API

biscuit-capi/src/lib.rs
```diff
@@ -0,0 +1,1251 @@
+use rand::prelude::*;
```
Geal, line 1: shouldn't this remove the file `biscuit-auth/src/capi.rs`?

Geal (changes requested):

biscuit-capi/src/lib.rs
ptondereau: Yes, it's just a copy paste from `biscuit-auth/src/capi.rs` and I just moved into the new crate. That's why I changed some visibility in some modules

biscuit-auth/src/token/mod.rs
```diff
@@ -221,6 +221,19 @@ impl Biscuit {
...
+        self.symbols.print_fact(fact)
+    }
+
+    pub fn print_rule(&self, rule: &crate::datalog::Rule) -> String {
+        self.symbols.print_rule(rule)
+    }
+
+    pub fn print_check(&self, check: &crate::datalog::Check) -> String {
+        self.symbols.print_check(check)
+    }
```
ptondereau, line 235: I don't know if it makes sense, maybe there is a better way

biscuit-auth/src/token/mod.rs
ptondereau: Addressed here: [`a71b2fb` (#234)](https://github.com/biscuit-auth/biscuit-rust/pull/234/commits/a71b2fb9413af1b07f176a5817785da32320b46a)

Thanks for the heads-up!

biscuit-auth/Cargo.toml
```diff
@@ -64,23 +63,6 @@ codspeed-bencher-compat = "2.6.0"
...
-  "Cargo.toml",
-  "cbindgen.toml",
-  "build.rs",
-  "examples/*.rs",
-  "LICENSE",
-  "README.md",
-  "src/*.rs",
-  "src/*/*.rs",
-  "tests/*.rs"
-]
```
ptondereau, line 82: I didn't backported this, if someone has a clue on what's the purpose of `include` config

biscuit-capi/src/lib.rs
Geal: could you delete `biscuit-auth/src/capi.rs` then? That way git will detect that the file moved and we'll be able to see the actual changes in the file

ptondereau: done sorry 🙇

ptondereau:
@geal or @divarvel can someone approve the workflow just to test the CI please?
Many thanks

ptondereau:
@Geal  tests green! 🎉

biscuit-auth/src/token/mod.rs
Geal: I think we should remove the `biscuit_block_fact`, `biscuit_block_rule` and `biscuit_block_check` functions from the C API. There is no equivalent in the rust API, because it does not make much sense to print facts or rules individually by their index. Instead there is `print_block_source` to show the datalog of an entire block

ptondereau: done here: [`00dfed0` (#234)](https://github.com/biscuit-auth/biscuit-rust/pull/234/commits/00dfed05c8cf2af9cb69052ba9b6436ff2169758)

Thanks for the suggest!

biscuit-auth/src/token/mod.rs
```diff
@@ -500,7 +500,7 @@ impl Biscuit {
         1 + self.blocks.len()
     }
 
-    pub(crate) fn block(&self, index: usize) -> Result<Block, error::Token> {
+    pub fn block(&self, index: usize) -> Result<Block, error::Token> {
```
Geal, line 503: now that the printing methods are updated, this change can be removed as well, because it would expose `token::block::Block`, which is not part of the public API (still marked as public in the code, but it was removed from the top level exports a while ago).
The `Biscuit::context` method should be more useful here: https://docs.rs/biscuit-auth/latest/biscuit_auth/struct.Biscuit.html#method.context

Geal: after that last change I think this will be ok to merge :)

ptondereau: and also `biscuit_block_count()`

Geal: probably not, it is better to align with the current rust API, where we encourage using the parsers and builders. The C API was designed along with the older rust API

ptondereau: Is this refactored example is valid?
```rust
#[no_mangle]
pub unsafe extern "C" fn biscuit_block_context(
    biscuit: Option<&Biscuit>,
    block_index: u32,
) -> *mut c_char {
    if biscuit.is_none() {
        update_last_error(Error::InvalidArgument);
        return std::ptr::null_mut();
    }

    let biscuit = biscuit.unwrap();

    let context = biscuit.0.context();

    match context.get(block_index as usize) {
        None => std::ptr::null_mut(),
        Some(context) => match CString::new(context.clone().unwrap()) {
            Ok(s) => s.into_raw(),
            Err(_) => {
                update_last_error(Error::InvalidArgument);
                std::ptr::null_mut()
            }
        },
    }
}
```

Geal: you'd want the following logic:
- if `get()` fails, you have the InvalidBlockId error
- if you get `Some(context)` but `context` is `None` (the block exists but had no context string, then return a null pointer without calling `update_last_error`

ptondereau: Ok I've removed `count()` method for now and update the `biscuit_block_context` methode here:

[`08f7900` (#234)](https://github.com/biscuit-auth/biscuit-rust/pull/234/commits/08f79006e062c840d4e7ca4baa15034081749bdf)

Geal (approved):
this is great, thanks :)

PR #237, opened 2024-11-03
do not build the C API on x86 OSX

Geal:


ptondereau (approved):

ptondereau:
We should also update the README that state we're building for macos x86

.github/workflows/release.yml
```diff
@@ -17,8 +17,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: macos-latest-large
-            target_arch: x86_64-apple-darwin
+          # - os: macos-latest-large
```
Geal, line 20:
```suggestion
          # deactivated for now, the x86 workers are more expensive
          # - os: macos-latest-large
```

PR #239, opened 2024-11-03
ci(capi): Add MUSL Support + SHA256 checksums

ptondereau:
- Add MUSL target
- Add SHA256 checksums
- rename release files to `biscuit-auth-capi-X.X.X-arch.tar.gz`
- download directly `cargo-c` instead of compiling
- Re-activate macos x86 build

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,30 +17,77 @@ jobs:
     strategy:
       matrix:
         include:
-          # deactivated for now, the x86 workers are more expensive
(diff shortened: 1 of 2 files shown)


Geal:
@ptondereau it looks like it will work. Do you feel there's anything to add?

ptondereau:
> @ptondereau it looks like it will work. Do you feel there's anything to add?

For the moment that's enough to start with I'll try to follow what has been done for the wasm plugin with C API then

PR #240, opened 2024-11-04
add samples for typeof on array / map

divarvel:


biscuit-auth/src/token/builder.rs
```diff
@@ -475,12 +475,10 @@ impl Term {
             Term::Map(m) => Term::Map(
                 m.into_iter()
                     .map(|(key, term)| {
-                        println!("will try to apply parameters on {key:?} -> {term:?}");
```
Geal, line 478: erf, sorry

Geal (approved):

PR #242, opened 2024-11-10
fix(ci): Fix CAPI build on macos

ptondereau:
## Purpose

- The build on macos-13 wasn't working because the pre-built `cargo-c` bins for macOS is only built for macos-arm architecture. We're now building both macOS arch on `macos-latest`
- I've spotted a bug where `inline-c` was always built in the release (see: https://github.com/biscuit-auth/biscuit-rust/actions/runs/11653423212/job/32445931933#step:6:94). We're now marking `inline-c` as a dev dependency and explicit requiring `libc` for `capi` build.
- Reworked the name of the CI jobs for ease of debugging: 
CleanShot 2024-11-10 at 14 51 52

I've tested the release workflow on my fork: https://github.com/ptondereau/biscuit-rust/actions/runs/11765631797 and https://github.com/ptondereau/biscuit-rust/releases/tag/capi-5.0.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: macos-13
+          - os: macos-latest
             target_arch: x86_64-apple-darwin
             arch: x86_64-macos
           - os: macos-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -86,7 +86,7 @@ jobs:
             shasum -a 256 "biscuit-auth-${{github.ref_name}}-${{matrix.arch}}.tar.gz" > "biscuit-auth-${{github.ref_name}}-${{matrix.arch}}.tar.gz.sha256"
           fi
       - name: Release
-        uses: softprops/action-gh-release@v1
+        uses: softprops/action-gh-release@v2
```
ptondereau, line 89: no change, only for removing deprecations notice

Geal (approved):

Issue #194, opened 2023-10-30
Hidden state in `Authorizer`


divarvel:
An `Authorizer` value wraps a `datalog::World` value, which contains facts and rules used for datalog evaluation. The `Authorizer` struct also contains an (optional) `Vec<Block>` representing a token (it is optional because an `Authorizer` can be created without a token. It also contains a `BlockBuilder` which contains facts, rules and checks added to the authorizer, as well as a `Vec<Policy>` containing policies added to the authorizer.

After calling `authorize()`, the `World` value contains facts from the token blocks, facts from the authorizer block builder and facts generated by rules.

The issue is that the `World` value does not always contain facts and rules from the authorizer builder:

- when adding a token to an authorizer, all facts and rules are loaded in the `World`
- when adding a fact or rule to an authorizer, it is added to the `BlockBuilder` but not the world
- when calling `authorize()`, facts and rules from the `BlockBuilder` are copied to the world, and datalog generation is run
- when calling `query` on `Authorizer`, datalog generation is automatically run on the current state of `World` (authorizer facts are not taken into account unless `authorize()` has been called first)

So, depending on whether `authorize()` and `query()` have been called, the authorizer world can be in various states:

- no authorizer facts or rules, no generated facts at all
- no authorizer facts or rules, but facts generated from the token contents
- authorizer facts or rules, and facts generated from both the token contents and the authorizer
- authorizer snapshots take a faithful copy of the inner `World` value, but only for facts. So taking a snapshot from an authorizer and then re-creating an authorizer from the snapshot will result in an authorizer with no authorizer rules in the datalog world

this makes relying on `World` contents tricky.

- `impl Display for Authorizer` only uses `World` for facts and rules currently -> this needs to be fixed (#195)
- `Authorizer.query()` only uses `World`, by design

Open questions:

- what should `query()` return? would it be possible to avoid running datalog generation in query if we can ensure it's done elsewhere?
- would it be possible to keep the `World` value in sync with the `BlockBuilder`. what would be the cost?
- does it make sense to continue adding facts and rules to an `Authorizer` after having called `authorize()`?

This affects #192 and #193

Issue #193, opened 2023-10-30
Non-deterministic output of `Authorizer.dump_code()`


divarvel:
Facts are stored as a `HashMap` internally, so their order is not defined.

The `Display` impl for `Authorizer` does sort everything, so it has a stable output. It also adds more information in comments (namely the facts and rules origins, which is extremely important).

I guess we could just call `.to_string()` in `dump_code()` to keep backwards compatibility. However, the `Display` implementation for `Authorizer` only uses facts and rules from the `Datalog::World` value, so it may discard facts and rules added to the authorizer before calling `authorize`. I guess that overlaps with #192

divarvel:
I am working on a fix for missing facts and rules

Issue #192, opened 2023-10-30
Can't query facts added with authorizer_merge!


Tipnos:
Below an example showing the issue:

```rust
use biscuit_auth::{macros::*, KeyPair};

fn main() {
    let mut authorizer = biscuit!(r#"fact("foo");"#)
        .build(&KeyPair::new())
        .unwrap()
        .authorizer()
        .unwrap();

    authorizer_merge!(&mut authorizer, r#"fact("bar");"#);

    let query_facts: Vec<(String,)> = authorizer.query(rule!("x($n) <- fact($n)")).unwrap();
    let (dumped_facts, _, _, _) = authorizer.dump();
    // [Fact { predicate: Predicate { name: "fact", terms: [Str("foo")] }, parameters: None },
    // Fact { predicate: Predicate { name: "fact", terms: [Str("bar")] }, parameters: Some({}) }]
    println!("{:?}", dumped_facts);

    // panic here!
    assert!(query_facts.len() == 2);
}

```
It looks like `query` don't lookup on facts added with `authorizer_merge!`

divarvel:
Thanks! I'm not sure how much of this is intended behaviour or not.

In any case, it should be properly documented.

wez:
I'm just getting my feet wet with this crate, so weight my comment accordingly.

I think the current behavior is desirable, and if different semantics are needed, I think an additional or alternative query method should be provided for that.

Here's my scenario:

The base token includes some facts (such as a list of resources that are to be updated as part of an operation) that I can trust, because they were put into the token by the original token issuer.

In my endpoint I plan to use the query method to extract those facts from the base token and use those to determine which resources my endpoint should operate on. This way my endpoint doesn't accept any additional parameters from the client; it will only operate on the trusted facts in the token.  Additional appended facts are not to be trusted; these are the same semantics as described and illustrated in https://doc.biscuitsec.org/reference/datalog#block-scoping

PR #250, opened 2024-11-27
authorizer builder

Geal:
Fix https://github.com/biscuit-auth/biscuit-rust/issues/194, https://github.com/biscuit-auth/biscuit-rust/issues/193, https://github.com/biscuit-auth/biscuit-rust/issues/192

This adds an `AuthorizerBuilder` struct that is used to create an `Authorizer`. All of the mutable behaviour, like adding facts or executing Datalog rules is moved into the builder, while the authorizer is limited to read-only queries (still requiring self mutability to track execution time). The `AuthorizerBuilder` is compatible with snapshots, to store and reuse checks and policies. It has a `build` method taking a token as argument, and a `build_unauthenticated` for authorization without token.

The builder APIs are alo changing. Before, we had the following:

```rust
 let mut builder = Biscuit::builder();	
builder.add_fact(r"right("file1", "read")"#)?;
builder.add_fact(r"right("file2", "read")"#)?;
let token = builder.build()?;
```

Builders are now constructed like this:

```rust
let token = Biscuit::builder()
    .fact(r"right("file1", "read")"#)?
    .fact(r"right("file2", "read")"#)?
    .build()?;
````

### DIFF
--- a/biscuit-auth/benches/token.rs
+++ b/biscuit-auth/benches/token.rs
@@ -16,11 +16,11 @@ fn create_block_1(b: &mut Bencher) {
     let root = KeyPair::new_with_rng(Algorithm::Ed25519, &mut rng);
 
     let token = Biscuit::builder()
-        .add_fact(fact("right", &[string("file1"), string("read")]))
+        .fact(fact("right", &[string("file1"), string("read")]))
         .unwrap()
-        .add_fact(fact("right", &[string("file2"), string("read")]))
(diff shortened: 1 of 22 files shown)


Geal:
the performance regression is due to `AuthorizerBuilder::build` executing Datalog rules before returning the authorizer, while the original benchmark was only testing the creation of an authorizer from a token, so it translates the block but does not execute rules

PR #251, opened 2024-11-27
refactor the builder API

Geal:


divarvel:
This should fix #194, #193, #192 and supersede #197

PR #252, opened 2024-11-29
disambiguate empty sets from empty maps

divarvel:
according to the grammar, `{}` could either mean *empty set* or *empty map*. Due to how the parser was written, it was always an empty set, and there was no way to have an empty map literal.

This is an issue because empty maps would still be displayed as `{}`, and maps have a different api than sets, resulting in evaluation errors.

This commit introduces `{,}` as the empty set literal, while `{}` is the empty map. The goal here is to be consistent with JSON, the reason why we chose the current syntax for arrays and maps.


divarvel:
coverage is OK, with tarpaulin run locally: 66.12% coverage, 5033/7612 lines covered, +0.54% change in coverage

i haven’t found the reason why tarpaulin fails in CI

biscuit-auth/src/datalog/mod.rs
```diff
@@ -743,12 +743,17 @@ pub struct RunLimits {
     pub max_time: Duration,
 }
 
+#[cfg(test)]
+const DEFAULT_DURATION_MILLIS: u64 = 1;
+#[cfg(not(test))]
+const DEFAULT_DURATION_MILLIS: u64 = 10_000;
```
Geal, line 749: shouldn't they be reversed? The longer one in tests?

Geal (approved):

biscuit-auth/src/datalog/mod.rs
divarvel: 🤦

divarvel: (i had removed the commit anyway but i’ll give another try)

PR #255, opened 2024-12-03
Authorizer API improvements

divarvel:
Based on tests with a private project:

- `impl Debug for Authorizer`
- allow constructing an `Authorizer` even if datalog evaluation fails

the second part is the biggest change, it makes running evaluation actually lazy (authorize and query functions ensure that evaluation is run). This allows constructing an Authorizer even if datalog evaluation fails. This is especially useful to take snapshots of failed datalog evaluations.

## TODO

- consider keeping track of failed evaluations (eg by replacing `Option<Duration>` with `NotStarted | Success(Duration) | Failed(Duration)` ) => maybe in a next version, this would require further changes to the snapshot data structure
- have a look at snapshot roundtrips => i have found an issue with `Authorizer::from_snapshot()`, but this will require extra work, in a separate PR

### DIFF
--- a/biscuit-auth/src/token/authorizer.rs
+++ b/biscuit-auth/src/token/authorizer.rs
@@ -31,10 +31,24 @@ pub struct Authorizer {
     pub(crate) blocks: Option<Vec<Block>>,
     pub(crate) public_key_to_block_id: HashMap<usize, Vec<usize>>,
     pub(crate) limits: AuthorizerLimits,
-    pub(crate) execution_time: Duration,
+    pub(crate) execution_time: Option<Duration>,
 }
 
 impl Authorizer {
+    pub fn run(&mut self) -> Result<Duration, error::Token> {
+        match self.execution_time {
+            Some(execution_time) => Ok(execution_time),
+            None => {
+                let start = Instant::now();
+                self.world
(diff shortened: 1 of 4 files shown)


biscuit-auth/src/token/authorizer.rs
```diff
@@ -31,10 +31,24 @@ pub struct Authorizer {
...
     pub(crate) public_key_to_block_id: HashMap<usize, Vec<usize>>,
     pub(crate) limits: AuthorizerLimits,
-    pub(crate) execution_time: Duration,
+    pub(crate) execution_time: Option<Duration>,
 }
 
 impl Authorizer {
+    pub fn run(&mut self) -> Result<Duration, error::Token> {
+        match self.execution_time {
+            Some(execution_time) => Ok(execution_time), // TODO should `run` be idempotent or raise an error if called twice?
```
divarvel, line 40: run being idempotent makes it really easy to lazily trigger evaluation in either authorize or query, so it is very important to have this ability. maybe as an internal function if we really want to make sure run triggers evaluation or fails if it can’t?

Geal: Yep it's good that this is idempotent

Geal (approved):

PR #265, opened 2025-02-04
Add pem/der loaders to PublicKey

Geal:
rebase of #212 over main

PEM key loading is only supported for ed25519 right now (so the relevant methods on the common KeyPair etc types are removed). Do we add it for p256 curves?


divarvel:
> PEM key loading is only supported for ed25519 right now (so the relevant methods on the common KeyPair etc types are removed). Do we add it for p256 curves?

yes, that would be great to have PEM support for both algorithms and for private and public keys

PR #272, opened 2025-02-23
Update README.md

CrypticError:
Fixed the example, previously it was referring to biscuit, the module, instead of biscuit2, the variable.


divarvel:
Thanks!

The failure on the codspeed benchmarks is unrelated, and benchmarks are not affected by your change anyways.

The only missing bit is the Eclipse eca check

divarvel:
Thank you!

PR #273, opened 2025-03-05
expose getter for `limits` on `AuthorizationBuilder`

divarvel:
The `limits` field used to be available on `Authorizer`, and is not visible in the builder


divarvel:
superseded by #274

PR #275, opened 2025-03-07
Add flexible string parsing for rule interpolation

jcorrv:
Previously, when using string interpolation to construct a rule like:

let component = "api_gateway".to_string();
rule(format!("right({}, $s) <- service($s)", component));

The parser would fail because it expected the string value to be explicitly quoted. This required cumbersome workarounds like manually adding quotes around interpolated values.

To maintain compatibility with other literals, the flexible parser excludes
 - booleans (true and false)
 - numeric literals (strings consisting only of digits and possibly a minus sign)
 - null
 - variables (strings starting with $)

 Testing:
 Adds two new tests in parser.rs and an example in testcases.rs

### DIFF
--- a/biscuit-auth/examples/testcases.rs
+++ b/biscuit-auth/examples/testcases.rs
@@ -143,6 +143,8 @@ fn run(target: String, root_key: Option<String>, test: bool, json: bool) {
 
     add_test_result(&mut results, third_party(&target, &root, test));
 
+    add_test_result(&mut results, third_party_macro(&target, &root, test));
+
     add_test_result(&mut results, check_all(&target, &root, test));
 
     add_test_result(&mut results, public_keys_interning(&target, &root, test));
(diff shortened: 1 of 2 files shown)


divarvel:
Hi, thanks for your contribution!

We _strongly_ advise against using regular string interpolation for building datalog (same reason as why string interpolation is advised against for SQL queries).

Thankfully, the library provides a way to properly interpolate values inside datalog fragments. There are two ways, depending if the datalog fragment is static, or dynamic.

If the datalog fragment is static, you can use the `rule!()` macro:

```rust
rule!(right({component}, $s) <- service($s));
```

see <https://docs.rs/biscuit-auth/latest/biscuit_auth/macros/macro.rule.html> for more details about the `rule!()` macro (other macros are available)

If the datalog fragment is dynamic, you have to handle interpolation yourself:

```rust
let r = rule("right({component}, $s) <- service($s)").unwrap().set("component", component).unwrap();
```

If possible, you should always strive for macros for two reasons:

- datalog validity is checked at compile-time
- parsing is done at compile time and the string is replaced by builder calls, so performance is better.

In any case, building datalog through string interpolation should be avoided and macros should be preferred, or at least manual parameter interpolation in other cases.

I will close the PR because we don’t want to support string interpolation, especially if that means weakening the parser.

Let me know if you need further help.

jcorrv:
Ah, that makes sense! Thanks for the detailed response @divarvel, I really appreciate it.

PR #276, opened 2025-03-09
use a different prefix for private keys than public keys

Geal:


divarvel:
Perfect, exactly what I had in mind. 

Build is failing because tests needs to be updated

Issue #290, opened 2025-06-09
GitHub action for release broken for Linux Arm64


AustinWise:
Triggering the release logic from the main branch fails because no build command is invoked. [See here](https://github.com/AustinWise/biscuit-rust/actions/runs/15525607534/job/43704838226) on my clone.

I tried changing the [this condition](https://github.com/eclipse-biscuit/biscuit-rust/blob/cea4b141cb82ef317a2262355d468f2eb540e4bc/.github/workflows/release.yml#L70) in release.yaml to cause the build to happen:

```diff
-        if: matrix.os == 'ubuntu-latest' && matrix.target_arch == 'x86_64-unknown-linux-gnu'
+        if: matrix.os == 'ubuntu-latest' && matrix.target_arch != 'x86_64-unknown-linux-musl'
```

However I then got this [error message](https://github.com/AustinWise/biscuit-rust/actions/runs/15524234533/job/43701536330):

> error: linking with `cc` failed: exit status: 1
>           /usr/bin/ld: /home/runner/work/biscuit-rust/biscuit-rust/target/aarch64-unknown-linux-gnu/release/deps/biscuit_auth.biscuit_auth.4ca180a0f95159a9-cgu.00.rcgu.o: Relocations in generic ELF (EM: 183)
>          /usr/bin/ld: /home/runner/work/biscuit-rust/biscuit-rust/target/aarch64-unknown-linux-gnu/release/deps/biscuit_auth.biscuit_auth.4ca180a0f95159a9-cgu.00.rcgu.o: error adding symbols: file in wrong format

If I understand correctly, this means that the linker tried to link object files from different architectures together.

I'm not familiar with `cargo-c`, so I'm not sure how to debug from here. This Linux arm64 logic was introduce in #278.

ptondereau:
Hello thanks for the report!
I'll provide a fix and for the error message, I think this is because the target should not be musl et also the target architecture émis wrong

PR #291, opened 2025-06-09
Add support for generating Windows release assets

AustinWise:
Three platforms are supported: x86_64, i686, and ARM64. There are the three user-land architectures supported by modern Windows ([arm32 was recently removed](https://learn.microsoft.com/en-us/windows/arm/arm32-to-arm64)). These platforms are supported by Rust as [tier 2 with host tools or above](https://doc.rust-lang.org/rustc/platform-support.html).

I've tested this PR in the following ways:

* trigged [a build](https://github.com/AustinWise/biscuit-rust/releases/tag/austin-test-11) and confirmed that all three architectures [generate assets](https://github.com/AustinWise/biscuit-rust/releases/tag/austin-test-11). 
* Downloaded the assets and confirmed that the `tar` command built into Windows 11 can extract the tar file.
* Used the x86_64, i686, and ARM64 DLL in my experimental [C# bindings](https://github.com/AustinWise/biscuit-csharp) to exercise several functions
* Statically linking on x86_64.

I ran into trouble with existing targets building. I'm pretty sure this is not caused by changes, see #290.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,6 +33,15 @@ jobs:
           - os: ubuntu-latest
             target_arch: aarch64-unknown-linux-gnu
             arch: arm64-linux
+          - os: windows-latest
+            target_arch: x86_64-pc-windows-msvc
+            arch: x86_64-windows
+          - os: windows-latest
+            target_arch: i686-pc-windows-msvc
+            arch: i686-windows
+          - os: windows-latest
+            target_arch: aarch64-pc-windows-msvc
+            arch: arm64-windows
 
     steps:
       - name: Checkout sources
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -77,13 +97,18 @@ jobs:
...
 
       - name: Compress
         run: tar cvzf biscuit-auth-${{github.ref_name}}-${{matrix.arch}}.tar.gz -C build/ .
 
       - name: Generate checksum
+        shell: bash
         run: |
-          if [[ "$RUNNER_OS" == "Linux" ]]; then
+          if [[ "$RUNNER_OS" == "Linux" || "$RUNNER_OS" == "Windows" ]]; then
             sha256sum "biscuit-auth-${{github.ref_name}}-${{matrix.arch}}.tar.gz" > "biscuit-auth-${{github.ref_name}}-${{matrix.arch}}.tar.gz.sha256"
```
ptondereau, line 112: Should we use .zip for windows instead? tar seems not installed by default?

.github/workflows/release.yml
```diff
@@ -62,12 +71,23 @@ jobs:
...
+        shell: cmd
+        run: |
+          curl -L %LINK%/%CARGO_C_VERSION%/%CARGO_C_FILE% -o %TEMP%\%CARGO_C_FILE%
+          tar xzf %TEMP%\%CARGO_C_FILE% -C %USERPROFILE%\.cargo\bin
+          del %TEMP%\%CARGO_C_FILE%
       - name: Run cargo-c tests
         run: cargo ctest --release --features="capi"
         if: matrix.os == 'ubuntu-latest' && matrix.target_arch == 'x86_64-unknown-linux-gnu'
 
       - name: Build C-API (Linux)
```
ptondereau, line 89: To ease the reading the CI maybe adding "arch" in the title

ptondereau (approved):
Thanks for this PR, I reviewed as I introduced this CI but the maintainers will have the final word.

About the error with linux + arm, I don't have such architecture at home but I will investigate

Geal:
@AustinWise thank you for looking into this. What are you using Biscuit on Windows for? Do you know that somebody is working on a .Net native implementation (if that could help). BTW to get this merged, you should fill the Eclipse Contributor Agreement, then once the cargo-c build on Linux ARM is solved it should be alright

AustinWise:
@Geal I have a project where I'm experimenting with Biscuits and sometimes I develop on a Windows computer. I don't have any grand plans at the moment for this project.

A native .NET implementation would be nice, but the [main Biscuit page](https://github.com/eclipse-biscuit/biscuit) says the .NET implementation only supports v1 of the specification. Is there a .NET implementation other than [dmunch/biscuit-net](https://github.com/dmunch/biscuit-net)?

I signed the Eclipse Contributor Agreement.

PR #292, opened 2025-06-09
Fix C-API release builds

ptondereau:
- We're upgrading `cargo-c` to `v0.10.13`
- We're fixing Linux arm64 build with additional toolchain libs.
- Added a cargo build profile just to remove debug symbol from the generated lib

### DIFF
--- a/biscuit-capi/Cargo.toml
+++ b/biscuit-capi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "biscuit-capi"
-version = "6.0.0-beta.3"                                    # Should keep the same version as biscuit-auth
+version = "6.0.0-beta.3"                                     # Should keep the same version as biscuit-auth
 description = "C API for Biscuit"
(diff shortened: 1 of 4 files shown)


ptondereau:
@divarvel need a review maybe before 6.0.0?

biscuit-capi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "biscuit-capi"
-version = "6.0.0-beta.3"                                    # Should keep the same version as biscuit-auth
+version = "6.0.0-beta.3"                                     # Should keep the same version as biscuit-auth
```
divarvel, line 3: regarding the version sync, does it still make sense, or could we decouple the two?

divarvel:
Do you want to issue a biscuit-c release at the same time as `biscuit-auth@6.0.0`?

ptondereau:
> Do you want to issue a biscuit-c release at the same time as `biscuit-auth@6.0.0`?

Yes if possible

PR #300, opened 2025-07-17
docs: add release checklist

divarvel:
based on the process for the first release review we went through for `biscuit-auth@6.0.0`


ptondereau:
FYI and if you're interested we use release-plz to handle the preparation and the release process. 
We have an example here: https://github.com/davidcole1340/ext-php-rs/blob/master/.release-plz.toml and see the GitHub action for the release

PR #306, opened 2025-08-12
fix: feature-gate `ToAnyParam` imports

divarvel:
`ToAnyParam` is only defined when `datalog-macro` is enabled, but some imports were not feature gated, preventing the library from compiling when the feature was not enabled (since it’s part of the default features, it was not common)

(tested with a crate, by setting `default-features = false`)


RemiBardon:
LGTM

PR #310, opened 2025-09-15
Scopes are deserialized twice

saoirse-a:


biscuit-auth/src/format/convert.rs
```diff
@@ -113,9 +113,6 @@ pub fn proto_block_to_token_block(
 
     detected_schema_version.check_compatibility(version)?;
 
-    let scopes: Result<Vec<Scope>, _> =
-        input.scope.iter().map(proto_scope_to_token_scope).collect();
-
```
divarvel, line 118: this will require removing the `?` below, i think, since this expression shadows the previous one.

PR #312, opened 2025-09-15
workspace: bump RustCrypto dependencies to new set of releases

baloo:
This is a very early attempt to update dependencies to use new releases of RustCrypto.

We (RustCrypto) don't have an final release yet, but this is meant to help test biscuit with code using pre-releases.

### DIFF
--- a/biscuit-auth/Cargo.toml
+++ b/biscuit-auth/Cargo.toml
@@ -26,8 +26,8 @@ uuid = ["dep:uuid"]
 pem = ["ed25519-dalek/pem", "ed25519-dalek/pkcs8"]
 
 [dependencies]
-rand_core = "^0.6"
-sha2 = "^0.9"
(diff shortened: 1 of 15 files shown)


divarvel:
This looks good! the failing check is because of a change in an error message, i’ll just have to refresh the samples

baloo:
Changes to `rand_core` in 0.9 and getrandom definitely affect wasm runtime. Cryptography changes, I don't expect any effect.

I haven't used wasm in a little while, and I haven't looked yet at the changes required there to get an RNG.

divarvel:
I think an update on #330 should supersede this PR

baloo:
Bumped the commit to the new set of releases.

I've also removed the direct dependencies on ecdsa and elliptic-curve as they are only use through p256. p256 should forward the features as expected.

PR #313, opened 2025-09-16
fix: make authorize_with_limits use provided limits for the run phase

### LINKED ISSUE #311: bug: `Authorizer::run()` behaves differently when using `Authorizer::authorize()` compared to `Authorizer::authorize_with_limits()`
When not using `.set_limits()` on an `AuthorizerBuilder`, the default limits are used. When you set limitd with `.set_limits()`, they are properly set on `self` (obviously).

https://github.com/eclipse-biscuit/biscuit-rust/blob/27ef8e36c4521c62bf0ef09875528f74dfac2449/biscuit-auth/src/token/builder/authorizer.rs#L264-L267

This works well with `Authorizer::authorize` because all limits are properly set beforehand:

https://github.com/eclipse-biscuit/biscuit-rust/blob/27ef8e36c4521c62bf0ef09875528f74dfac2449/biscuit-auth/src/token/authorizer.rs#L365-L375

And the call to `self.run()` uses the limits from `self` properly.

---

Now, when I use `Authorizer::authorize_with_limits`, the call to `self.run()` does not propagate these limits as `run` does not take any limits:

https://github.com/eclipse-biscuit/biscuit-rust/blob/27ef8e36c4521c62bf0ef09875528f74dfac2449/biscuit-auth/src/token/authorizer.rs#L42-L54

Hence, the limits from `self` are used - **which are not the limits we explicitly provided when calling `authorize_with_limits()`**:

https://github.com/eclipse-biscuit/biscuit-rust/blob/27ef8e36c4521c62bf0ef09875528f74dfac2449/biscuit-auth/src/token/authorizer.rs#L48

---

This left me with weird test failures because `authorize_with_limits()` does not propagate the limits properly to `run()`.

divarvel:
- expose `Authorizer::run_with_limits`
- use it in `Authorizer::authorize_with_limits`

fixes #311

The provided limits were correctly used in the authorization step, but not in the datalog fact generation step. For authorizers where the run step was already done, it did not change anything, but for authorizers not run yet, the provided limits were not used.


georglauterbach (approved):
LGTM 👌🏽

PR #314, opened 2025-09-16
WIP: remove `ToAnyParam` and use `Into<Param>` instead

divarvel:
Continuation of #302 

The goal is to:

- remove the possibility of passing a public key where a term is expected
- remove `ToAnyParam` in favor of `Into<AnyParam>` for simplicity
- (possibly) extend the passing of public keys to support multiple keys at once, and keywords.

The main difference is that `ToAnyParam` takes a reference, while `Into` consumes its input.

# todo

- [x] try using `Into<Term>` instead of `T: ToAnyParam`
- [ ] try using `Into<(PublicKey, Vec<PublicKey>)>` instead of `PublicKey`

### DIFF
--- a/biscuit-auth/src/token/builder.rs
+++ b/biscuit-auth/src/token/builder.rs
@@ -173,13 +173,6 @@ pub trait ToAnyParam {
     fn to_any_param(&self) -> AnyParam;
 }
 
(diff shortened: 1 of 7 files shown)


Geal (approved):

divarvel:
Superseded by #334

PR #320, opened 2026-03-02
Add aws-lc-rs crypto backend with trait-based abstraction

preuss-adam:
## Summary

- Add trait-based crypto backend abstraction (Backend, KeyPairImpl, PrivateKeyImpl, PublicKeyImpl, KeySerialization, KeyPemSerialization)
- Move existing ed25519-dalek/p256 code into `default/` backend implementing the new traits
- Add aws-lc-rs backend with Ed25519 and secp256r1 support
- Feature-gate backends: `default-backend` (ed25519-dalek/p256) and `awslc-backend` (aws-lc-rs)
- Remove unused deps: sha2, prost-types, getrandom, ecdsa, elliptic-curve
- Upgrade pkcs8 to 0.10.2

### Ed25519 verification (third-party checks)

| Blocks | default-backend | awslc-backend | Speedup |
|--------|----------------|---------------|---------|
| 0 | 85us | 56us | 1.5x |
| 1 | 108us | 78us | 1.4x |
| 2 | 167us | 113us | 1.5x |
| 3 | 197us | 148us | 1.3x |
| 4 | 241us | 185us | 1.3x |
| 5 | 285us | 222us | 1.3x |
| 6 | 337us | 253us | 1.3x |
| 7 | 374us | 289us | 1.3x |

### P256 verification (third-party checks)

| Blocks | default-backend | awslc-backend | Speedup |
|--------|----------------|---------------|---------|
| 0 | 406us | 76us | 5.3x |
| 1 | 731us | 127us | 5.8x |
| 2 | 975us | 185us | 5.3x |
| 3 | 1,253us | 230us | 5.4x |
| 4 | 1,561us | 284us | 5.5x |
| 5 | 1,847us | 338us | 5.5x |
| 6 | 2,091us | 393us | 5.3x |
| 7 | 2,445us | 446us | 5.5x |

## Usage

Default (ed25519-dalek + p256):
```toml
biscuit-auth = "6.0.0"
```

aws-lc-rs backend:
```toml
biscuit-auth = { version = "6.0.0", default-features = false, features = ["awslc-backend", "pem", "regex-full", "datalog-macro"] }
```


divarvel:
hi, as an update, i still plan on reviewing this; thanks for your patience.

divarvel (commented):
One of the pending projects that’s related to this effort is allowing the use of a KMS, same as what was done in biscuit-java https://github.com/eclipse-biscuit/biscuit-java/pull/113

I think the traits, as they are designed, would work, but the naming would be a bit misleading, as they focus on abstracting over things (public key, private key) instead of the actual behavior (signing, verifying).
The traits as they are defined have a granularity that would fit the bill (especially separating the key serialization functions), but the traits constraints as they stand would not allow using a KMS for the authority key and a regular keypair for attenuation keys. I feel that would be resolved by adding an equivalent of `BiscuitBuilder::build` that would be generic on the signer argument (maybe with a default defined with cfg macros).

Finally, and that’s really nitpicking, I was a bit taken aback by the `xxImpl` naming for a trait.

I think we should design things around two main traits:

`Signer`, used in `BiscuitBuilder::build()` and `ThirdPartyRequest::create_block()`. types implementing it are able to sign a payload, expose a public key and declare their algorithm.
`KeyPair`, used in `Biscuit::append()`. Types implementing it can be generated, expose a public and private key that can be serialized / deserialized and declare their algorithm.

Those two traits would handle the places where the biscuit impl itself needs to call out to the crypto layer. For the rest, there is no strong need to be generic, since it’s code written by library users (still, defining a trait would make things a bit nicer to user by enforcing consistency between implementations).
imo, moving parametricity to actual call sites in the biscuit library would let us get on by with smaller traits and simpler code, and default generic types defined through `cfg` attributes would keep the nice property of deciding which implementation to use through cargo features, while keeping the opportunity to strategically override things

divarvel:
I think this is superseded by #334

PR #322, opened 2026-04-09
Remove dead code in crypto and fix algo bug

saoirse-a:
The (ultimately not public) ed25519 and p256 crypto types expose a lot of methods that are never called; dead code left over from when the abstraction over multiple signing algorithms was added.

The P256 code in several places had copy pasted code from ed25519 which incorrectly specified the algorithm as as ed25519; most of these were dead code, but one was in the Hash implementation of P256's public key, which incorrectly hashed the ed25519 identifier.

### DIFF
--- a/biscuit-auth/src/crypto/p256.rs
+++ b/biscuit-auth/src/crypto/p256.rs
@@ -3,13 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 #![allow(non_snake_case)]
-use crate::{error::Format, format::schema};
+use crate::error::Format;
 
 use super::error;
 use super::Signature;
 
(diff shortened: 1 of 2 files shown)


biscuit-auth/src/crypto/p256.rs
```diff
@@ -326,18 +286,20 @@ impl PublicKey {
 
 impl Hash for PublicKey {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        (crate::format::schema::public_key::Algorithm::Ed25519 as i32).hash(state);
+        (crate::format::schema::public_key::Algorithm::Secp256r1 as i32).hash(state);
```
saoirse-a, line 289: highlighting this as the actual bug fix

divarvel (approved):

PR #323, opened 2026-04-09
Comment out println in rule application

saoirse-a:
Perhaps this should be raised as an error of some kind, but it certainly shouldn't be printed to stdout.

### DIFF
--- a/biscuit-auth/src/datalog/mod.rs
+++ b/biscuit-auth/src/datalog/mod.rs
@@ -175,7 +175,7 @@ impl Rule {
                             Term::Variable(i) => match h.get(i) {
                               Some(val) => p.terms[index] = val.clone(),
                               None => {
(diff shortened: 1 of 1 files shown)


divarvel:
I think it should be an error indeed. Maybe as a follow-up task?

biscuit-auth/src/datalog/mod.rs
```diff
@@ -175,7 +175,7 @@ impl Rule {
                             Term::Variable(i) => match h.get(i) {
                               Some(val) => p.terms[index] = val.clone(),
                               None => {
-                                println!("error: variables that appear in the head should appear in the body and constraints as well");
+                                //println!("error: variables that appear in the head should appear in the body and constraints as well");
```
divarvel:
```suggestion
                                // head variables should be bound in the body predicates
```

divarvel (approved):

PR #324, opened 2026-04-10
Eliminate unnecessary cloning in any/all loop

saoirse-a:
The RHS expression only needs to be constructed once, not repeatedly inside the loop, because evaluating it does not consume it.


divarvel (approved):
Neat! How did you notice it?

saoirse-a:
I asked Claude to identify low hanging fruit for optimising the authorisation logic; it identifies this issue as well as #325 (and also several non-issues naturally) and then I implemented these changes by hand.

PR #326, opened 2026-04-16
Resolve clippy and fmt issues

saoirse-a:
This fixes issues identified by clippy - the most common was taking a reference to something that was already a reference.

A few warnings are supressed rather than fixed, mostly because fixing them could be a breaking change.

Also runs cargo fmt over the project.

### DIFF
--- a/biscuit-auth/benches/token.rs
+++ b/biscuit-auth/benches/token.rs
@@ -72,7 +72,7 @@ fn append_block_2(b: &mut Bencher) {
     b.bytes = (data.len() - base_data.len()) as u64;
     assert_eq!(b.bytes, 189);
     b.iter(|| {
-        let token = Biscuit::from(&base_data, &root.public()).unwrap();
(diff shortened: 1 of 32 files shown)


divarvel:
This generally looks good to me. I know @Geal was not keen on having rustfmt early on, but it’s now common enough that it would be good now.

if we reformat the codebase we should also add a CI check as well imo.

saoirse-a:
The format changes were pretty much all in schema.rs; codebase seems to already have had rustfmt run over it at some point

saoirse-a:
That's actually also what's making the tests fail (its the test that diffs the schema with a generated schema), so if we want to do that we need to also update the test to format the code it generates

divarvel:
> That's actually also what's making the tests fail (its the test that diffs the schema with a generated schema), so if we want to do that we need to also update the test to format the code it generates

i’d rather exclude generated files from rustfmt

saoirse-a:
Removed the format changes from the schema

divarvel (approved):

PR #327, opened 2026-04-16
[BREAKING] Fix TrustedOrigins default issue

saoirse-a:
TrustedOrigins implements an inherent constructor called default and also has a derived implementation of Default. **These two constructors have different behavior.**

This means that `TrustedOrigins::default()` and `<TrustedOrigins as Default>::default()` construct different `TrustedOrigins` types. This is surprising.

This PR changes the behavior of the Default trait constructor to match the behavior of the inherent constructor.

### DIFF
--- a/biscuit-auth/CHANGELOG.md
+++ b/biscuit-auth/CHANGELOG.md
@@ -1,3 +1,7 @@
+# `7.0.0`
+
+- `TrustedOrigin`'s `Default` implementation changed to match its inherent `default` function (#327)
+
 # `6.0.0`
 
 - support for `pem` / `der` private and public keys (#212 and #265)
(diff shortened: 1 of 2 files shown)


divarvel (approved):
oh, nice catch

divarvel:
Could you add an entry in the changelog as well?

saoirse-a:
Added to the changelog under 7.0.0

Geal (approved):

divarvel:
Since #334 is moving on, I guess we have a strong case for a major release

divarvel:
@saoirse-a I’ll merge #334 first, can you rebase this one on top of main afterwards? I’ll merge it then

saoirse-a:
rebased

saoirse-a:
Actually I also updated it to delete the inherent `default` constructor; since `Default` is in the prelude this shouldn't break anything, there's no reason to have it.

PR #330, opened 2026-05-15
Bump dependencies

avandecreme:
I couldn't bump `rand_core` and `rand` because they are tied to the version used by `ed25519-dalek`.

### DIFF
--- a/biscuit-auth/src/error.rs
+++ b/biscuit-auth/src/error.rs
@@ -67,10 +67,11 @@ impl From<base64::DecodeError> for Token {
             base64::DecodeError::InvalidByte(offset, byte) => {
                 Base64Error::InvalidByte(offset, byte)
             }
-            base64::DecodeError::InvalidLength => Base64Error::InvalidLength,
+            base64::DecodeError::InvalidLength(len) => Base64Error::InvalidLength(len),
             base64::DecodeError::InvalidLastSymbol(offset, byte) => {
                 Base64Error::InvalidLastSymbol(offset, byte)
             }
+            base64::DecodeError::InvalidPadding => Base64Error::InvalidPadding,
         };
 
         Token::Base64(err)
@@ -90,8 +91,9 @@ impl From<Execution> for Token {
 #[cfg_attr(feature = "serde-error", derive(serde::Serialize, serde::Deserialize))]
 pub enum Base64Error {
     InvalidByte(usize, u8),
-    InvalidLength,
+    InvalidLength(usize),
     InvalidLastSymbol(usize, u8),
+    InvalidPadding,
 }
 
 impl std::fmt::Display for Base64Error {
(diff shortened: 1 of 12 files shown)


biscuit-auth/src/error.rs
```diff
@@ -90,8 +91,9 @@ impl From<Execution> for Token {
 #[cfg_attr(feature = "serde-error", derive(serde::Serialize, serde::Deserialize))]
 pub enum Base64Error {
     InvalidByte(usize, u8),
-    InvalidLength,
+    InvalidLength(usize),
     InvalidLastSymbol(usize, u8),
+    InvalidPadding,
```
avandecreme, line 96: This is a breaking change

biscuit-auth/src/format/convert.rs
```diff
@@ -710,78 +714,78 @@ fn proto_op_to_token_op(op: &schema::Op) -> Result<Op, error::Format> {
...
+                (Ok(_), Some(_)) => {
                     return Err(error::Format::DeserializationError(
                         "deserialization error: ffi name set on a regular unary operation"
                             .to_string(),
                     ))
                 }
-                (None, _) => {
+                (Err(_), _) => {
                     return Err(error::Format::DeserializationError(
                         "deserialization error: unary operation is empty".to_string(),
```
avandecreme, line 736: Not sure about that error message

biscuit-auth/src/format/convert.rs
```diff
@@ -710,78 +714,78 @@ fn proto_op_to_token_op(op: &schema::Op) -> Result<Op, error::Format> {
...
                         "deserialization error: ffi name set on a regular binary operation"
                             .to_string(),
                     ))
                 }
-                (Some(op_binary::Kind::TryOr), None) => Op::Binary(Binary::TryOr),
-                (None, _) => {
+                (Ok(op_binary::Kind::TryOr), None) => Op::Binary(Binary::TryOr),
+                (Err(_), _) => {
                     return Err(error::Format::DeserializationError(
                         "deserialization error: binary operation is empty".to_string(),
```
avandecreme, line 790: Same

biscuit-quote/Cargo.toml
```diff
@@ -13,7 +13,7 @@ proc-macro = true
 biscuit-parser = { path = "../biscuit-parser", features = ["datalog-macro"], version = "0.2.0" }
 proc-macro2 = "1"
 quote = "1.0.14"
-syn = { version = "1.0.85", features = ["full", "extra-traits"] }
+syn = { version = "2.0.117", default-features = false }
```
avandecreme, line 16: syn 3 was released recently, I guess we could skip version 2

avandecreme:
@divarvel any interest in this?

ixcans:
Hi @avandecreme Thanks for the PR!


The blocker in the description looks like it has cleared since this was opened in May: ed25519-dalek 3.0.0 went stable on 2026-07-06 and takes rand_core ^0.10, so rand/rand_core are no longer pinned back by it.

For what it's worth, the rest of the stack this PR would need has landed as stable releases too, all after this PR was opened:

 
 
crate | stable version | released
-- | -- | --
signature | 3.0.0 | 2026-05-02
p256 | 0.14.0 | 2026-07-03
curve25519-dalek | 5.0.0 | 2026-07-06
ed25519-dalek | 3.0.0 | 2026-07-06

 

That may also unblock #312, its description notes there was no final RustCrypto release at the time, and that is no longer the case.

A downstream data point in case it helps prioritise: we build services on biscuit-auth, and because our own code shares SigningKey / PublicKey types with the library across an API boundary, we can't dual-version the crypto stack. So biscuit-auth 6.0.0's requirements transitively hold our whole workspace at ed25519-dalek 2, p256 0.13, rand 0.8, rand_core 0.6 and getrandom 0.2, we attempted the upgrade this week and had to back all of it out, while every other dependency moved.


Hi @divarvel , when you get a chance, could you please give a review to this PR? Thank you in advance!

PR #331, opened 2026-05-28
[BREAKING] Separate biscuit-proto to own crate

saoirse-a:
Instead of exposing the generated protobuf code in biscuit-auth, this PR introduces a `biscuit-proto` crate which exposes the protobuf types. No types from biscuit-proto appear in the public API of biscuit-auth, so the generated code can evolve independently of biscuit-auth without breakage. For example, if tonic changes its codegen or a change to the protobuf format would result in a breaking change to the Rust types, this doesn't require a breaking update to biscuit-auth.

Practically, this means:

1. No public conversion methods between the biscuit-auth types and the proto types; instead they can be serialised to bytes and deserialised as the proto types if needed.
2. Consistently use the `Algorithm` type defined in biscuit-auth, instead of the proto generated one; this is mostly a code clean up for some modules where both types were intermixed.
3. Add several additional `dump_*` methods to `Authorizer` to allow the test cases example dump the authorizer's content without converting it to a snapshot.

### Future work

As a follow on change, I want to refactor the samples generation to separate testing and sample generation, with sample generation as a separate binary crate (not uploaded to crates.io). Then biscuit-auth's sample tests can consume the checked in samples in the same way as the other language libraries, instead of being the source of truth for the samples (which will instead be the new `biscuit-generate-samples` crate.)


divarvel:
LGTM

divarvel (approved):

divarvel:
Now that #334 is merged, this could go in imo (once rebased)

divarvel:
The coverage test was a timeout, but there’s something weird going on with the samples

PR #334, opened 2026-06-24
[BREAKING] Abstract biscuit-auth over crypto implementation

saoirse-a:
This adds traits to the biscuit-auth public interface for cryptographic operations and abstracts the Biscuit and related types over those traits.

The traits added are `Sign`, `Verify`, `SerializePublicKey` and `SerializePrivateKey`. Together, these represent the necessary behavior for public and private keys, distinguishing those which can be serialised into the token and those which can't:

```rust
pub trait Sign {
    type PublicKey: Verify;

    fn sign(&self, data: &[u8]) -> Result<Signature, error::Format>;
    fn public(&self) -> Self::PublicKey;
    fn algorithm(&self) -> Algorithm;
}

pub trait Verify {
    fn verify_signature(
        &self,
        data: &[u8],
        signature: &Signature,
    ) -> Result<(), error::Format>;
    fn algorithm(&self) -> Algorithm;
}

pub trait SerializePrivateKey: Sign<PublicKey: SerializePublicKey> + Clone + Sized {
    fn new_with_rng<R: RngCore + CryptoRng>(algorithm: Algorithm, rng: &mut R) -> Self;
    fn from_bytes_and_algorithm(algorithm: Algorithm, bytes: &[u8]) -> Result<Self, error::Format>;
    fn to_bytes(&self) -> Zeroizing<Vec<u8>>;
}

pub trait SerializePublicKey: Verify + Clone + PartialEq + Sized {
    fn from_bytes_and_algorithm(algorithm: Algorithm, bytes: &[u8]) -> Result<Self, error::Format>;
    fn to_bytes(&self) -> Vec<u8>;
}
```

The `Biscuit` type is parameterised by a `SerializePrivateKey`, which is the type of the key contain in its proof. All of the public keys in the token are the associated serialisable public key type for that private key.

The issuer key and third party signing keys can be different types from the key contained in the token; for the issuer neither the private key nor the public key need be serialisable, for the third party key only the public key needs to be serialisable because it appears in the token.

The existing `PublicKey` and `PrivateKey` types implement these traits; no other implementations are added directly to biscuit-auth with this PR (we could either add implementations for other crypto libraries like aws-lc-rs, maybe feature gated, or just require users to write their own implementations).

To support this change, two significant refactors were performed:

1. The `Keypair` type was eliminated. This was effectively redundant with the `PrivateKey` type, removing it made the set of traits required simpler.
2. The `PublicKeys` table used in authorization execution now stores an inert representation of the public key instead of the `PublicKey` type, because it doesn't need to verify signatures, just compare keys for equality. This avoids more abstraction points over crypto types in the Datalog code.

### DIFF
--- a/biscuit-auth/CHANGELOG.md
+++ b/biscuit-auth/CHANGELOG.md
@@ -1,3 +1,7 @@
+# `7.0.0`
+
+- Abstract biscuits over the crypto implementation (#334)
+
 # `6.0.0`
 
 - support for `pem` / `der` private and public keys (#212 and #265)
--- a/biscuit-auth/CHANGELOG.md
+++ b/biscuit-auth/CHANGELOG.md
@@ -1,6 +1,24 @@
 # `7.0.0`
 
-- Abstract biscuits over the crypto implementation (#334)
+- Abstract biscuits over the crypto implementation (#334) Tokens are generic over the key type used
+  to sign them and four new traits are added: `Sign` and `SerializePrivateKey` for private keys and
+  `Verify` and `SerializePublicKey` for public keys.
+- New `PublicKeyData` type representing the inert data representation of a public key as it is
+  stored in a token, as well as `PublicKeys`, the block's public key table.  (#334)
+
+## Breaking changes
+
+- `KeyPair` is removed; use `PrivateKey` directly instead of `KeyPair`. APIs that previously took a
+  keypair are replaced with APIs that take a `PrivateKey` (i.e. `append_with_keypair` ->
+  `append_with_key`, `append_third_party_with_keypair` -> `append_third_party_with_key`).
+- `Biscuit` and `UnverifiedBiscuit` are generic over the private key type contained in their
+  proof. These are defaulted to `PrivateKey`, so this change should be transparent for most users
+  who use the default crypto types.
+- `RootKeyProvider` has an associated `Key` type, which is the type of the root key that is
+  provided.
+- All `algorithm()` methods return `builder::Algorithm` instead of
+  `format::schema::public_key::Algorithm`.
+- Datalog scopes now take `PublicKeyData` instead of the cryptographic `PublicKey` type.
 
 # `6.0.0`
 
--- a/biscuit-auth/CHANGELOG.md
+++ b/biscuit-auth/CHANGELOG.md
@@ -5,6 +5,9 @@
   `Verify` and `SerializePublicKey` for public keys.
 - New `PublicKeyData` type representing the inert data representation of a public key as it is
   stored in a token, as well as `PublicKeys`, the block's public key table.  (#334)
+- `Term` implements `From<Uuid>` with the `uuid` feature.
+- Scope parameters are `Into<PublicKeyData>` so they can take anything which can be converted into a
+  public key (including `&K: SerializePublicKey`).
 
 ## Breaking changes
 
@@ -19,6 +22,10 @@
 - All `algorithm()` methods return `builder::Algorithm` instead of
   `format::schema::public_key::Algorithm`.
 - Datalog scopes now take `PublicKeyData` instead of the cryptographic `PublicKey` type.
+- The `ToAnyParam` trait and `AnyParam` enum are removed; users who were implementing `ToAnyParam`
+  for custom types should implement `From<MyType> for Term` instead.
+- The `set_macro_param` and `set_macro_scope_param` methods are removed; macros now call
+  `set_lenient` and `set_scope_lenient`.
 
 # `6.0.0`
 
--- a/biscuit-auth/src/crypto/mod.rs
+++ b/biscuit-auth/src/crypto/mod.rs
@@ -11,19 +11,25 @@
 //!
 //! The implementation is based on [ed25519_dalek](https://github.com/dalek-cryptography/ed25519-dalek).
 #![allow(non_snake_case)]
+mod ed25519;
+mod p256;
+mod traits;
+
+use std::fmt;
+use std::hash::Hash;
+use std::str::FromStr;
+
+use nom::Finish;
+use rand_core::{CryptoRng, RngCore};
+use zeroize::Zeroizing;
+
 use crate::builder::Algorithm;
 use crate::format::schema;
 use crate::format::ThirdPartyVerificationMode;
 
 use super::error;
-mod ed25519;
-mod p256;
 
-use nom::Finish;
-use rand_core::{CryptoRng, RngCore};
-use std::fmt;
-use std::hash::Hash;
-use std::str::FromStr;
+pub use self::traits::{Verify, Sign, SerializePublicKey, SerializePrivateKey};
 
 /// the private part of a signing key pair
 #[derive(Debug, Clone, PartialEq)]
@@ -116,13 +122,6 @@ impl PrivateKey {
         }
     }
 
-    pub fn private(&self) -> PrivateKey {
-        match self {
-            PrivateKey::Ed25519(key) => PrivateKey::Ed25519(key.clone()),
-            PrivateKey::P256(key) => PrivateKey::P256(key.clone()),
-        }
-    }
-
     pub fn public(&self) -> PublicKey {
         match self {
             PrivateKey::Ed25519(key) => PublicKey::Ed25519(key.public()),
@@ -130,10 +129,10 @@ impl PrivateKey {
         }
     }
 
-    pub fn algorithm(&self) -> crate::format::schema::public_key::Algorithm {
+    pub fn algorithm(&self) -> Algorithm {
         match self {
-            PrivateKey::Ed25519(_) => crate::format::schema::public_key::Algorithm::Ed25519,
-            PrivateKey::P256(_) => crate::format::schema::public_key::Algorithm::Secp256r1,
+            PrivateKey::Ed25519(_) => Algorithm::Ed25519,
+            PrivateKey::P256(_) => Algorithm::Secp256r1,
         }
     }
 
@@ -153,8 +152,8 @@ impl PrivateKey {
     /// serializes to an hex-encoded string, prefixed with the key algorithm
     pub fn to_prefixed_string(&self) -> String {
         let algorithm = match self.algorithm() {
-            schema::public_key::Algorithm::Ed25519 => "ed25519-private",
-            schema::public_key::Algorithm::Secp256r1 => "secp256r1-private",
+            Algorithm::Ed25519 => "ed25519-private",
+            Algorithm::Secp256r1 => "secp256r1-private",
         };
         format!("{algorithm}/{}", self.to_bytes_hex())
     }
@@ -219,7 +218,6 @@ impl PrivateKey {
     }
 }
 
-
 impl Default for PrivateKey {
     fn default() -> Self {
         Self::new()
@@ -239,6 +237,39 @@ impl FromStr for PrivateKey {
                 "Missing key algorithm".to_string(),
             )),
         }
+    }
+}
+
+impl Sign for PrivateKey {
+    type PublicKey = PublicKey;
+
+    fn sign(&self, data: &[u8]) -> Result<Signature, error::Format> {
+        self.sign(data)
+    }
+
+    fn public(&self) -> Self::PublicKey {
+        self.public()
+    }
+
+    fn algorithm(&self) -> Algorithm {
+        match self {
+            PrivateKey::Ed25519(_) => Algorithm::Ed25519,
+            PrivateKey::P256(_) => Algorithm::Secp256r1,
+        }
+    }
+}
+
+impl SerializePrivateKey for PrivateKey {
+    fn new_with_rng<R: RngCore + CryptoRng>(algorithm: Algorithm, rng: &mut R) -> Self {
+        Self::new_with_rng(algorithm, rng)
+    }
+
+    fn from_bytes_and_algorithm(algorithm: Algorithm, bytes: &[u8]) -> Result<Self, error::Format> {
+        Self::from_bytes(bytes, algorithm)
+    }
+
+    fn to_bytes(&self) -> Zeroizing<Vec<u8>> {
+        self.to_bytes()
     }
 }
 
@@ -355,10 +386,10 @@ impl PublicKey {
         }
     }
 
-    pub fn algorithm(&self) -> crate::format::schema::public_key::Algorithm {
+    pub fn algorithm(&self) -> Algorithm {
         match self {
-            PublicKey::Ed25519(_) => crate::format::schema::public_key::Algorithm::Ed25519,
-            PublicKey::P256(_) => crate::format::schema::public_key::Algorithm::Secp256r1,
+            PublicKey::Ed25519(_) => Algorithm::Ed25519,
+            PublicKey::P256(_) => Algorithm::Secp256r1,
         }
     }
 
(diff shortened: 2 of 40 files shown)


saoirse-a:
I think there is more work to be done to fully integrate the traits:

1. Some APIs in the builders aren't generic over crypto providers still.
2. Some of the Datalog auth code (like the `Authorizer` itself) is abstract over key types because it stores more data from the token than it really needs to.

Before I go through all of that with a fine tooth comb I wanted to get confirmation that this is the right direction.

divarvel:
I like this approach as it neatly handles both KMS and pluggable crypto providers.
It also opens the door to unifying Biscuit and UnverifiedBiscuit (I think)

saoirse-a:
Updated, I think this is ready now.

biscuit-auth/tests/macros.rs
```diff
@@ -122,12 +123,10 @@ fn authorizer_macro_trailing_comma() {
...
 #[test]
 fn biscuit_macro() {
-    use biscuit_auth::PublicKey;
     let pubkey = PublicKey::from_bytes(
-        &hex::decode("6e9e6d5a75cf0c0e87ec1256b4dfed0ca3ba452912d213fcc70f8516583db9db").unwrap(),
         biscuit_auth::builder::Algorithm::Ed25519,
-    )
-    .unwrap();
+        hex::decode("6e9e6d5a75cf0c0e87ec1256b4dfed0ca3ba452912d213fcc70f8516583db9db").unwrap(),
+    );
```
divarvel: nit: it would make sense to use the from_str implementations here now that we have those (not directly related to the PR, but since we have to update these lines)

biscuit-auth/src/token/public_keys.rs
```diff
@@ -3,77 +3,145 @@
...
             None => {
-                self.keys.push(*k);
+                self.keys.push(k.clone());
                 (self.keys.len() - 1) as u64
             }
         }
     }
 
-    pub fn insert_fallible(&mut self, k: &PublicKey) -> Result<u64, error::Format> {
+    pub(crate) fn insert_serialize<K: SerializePublicKey>(&mut self, k: &K) -> u64 {
```
divarvel: nit: `insert_serialize` makes it sound like `serialize` is part of the action being performed, whereas it’s because of the trait name. Maybe making the trait name more explicit would help?

biscuit-auth/src/token/public_keys.rs
```diff
@@ -3,77 +3,145 @@
...
-                Ok((self.keys.len() - 1) as u64)
+                self.keys.push(k.clone());
+                (self.keys.len() - 1) as u64
             }
         }
     }
 
-    pub fn get(&self, k: &PublicKey) -> Option<u64> {
-        self.keys.iter().position(|key| key == k).map(|i| i as u64)
+    pub(crate) fn insert_proto(&mut self, k: &schema::PublicKey) -> u64 {
```
divarvel: i’m not sure having a dependency to the proto module makes sense here. i’d rather have all the dependencies to proto be located in `convert`

biscuit-auth/src/token/authorizer.rs
```diff
@@ -917,10 +918,10 @@ mod tests {
     use token::builder::{self, load_and_translate_block, var};
     use token::{public_keys::PublicKeys, DATALOG_3_1};
 
-    use crate::PublicKey;
+    use crate::token::public_keys::PublicKey as InertPublicKey;
```
divarvel: what’s the intended meaning of `Inert` here?

biscuit-auth/src/crypto/traits.rs
```diff
@@ -0,0 +1,34 @@
+use rand::{CryptoRng, RngCore};
+use zeroize::Zeroizing;
+
+use crate::builder::Algorithm;
+use crate::crypto::Signature;
+use crate::error;
+
+pub trait Verify {
```
divarvel, line 8: nit: the trait names make sense in the context of this trait, but i’ve found those to be a bit harder to understand when reviewing call sites.

biscuit-auth/src/token/builder/authorizer.rs
```diff
@@ -158,7 +161,7 @@ impl AuthorizerBuilder {
                 res?;
             }
             for (name, value) in &scope_params {
-                let res = match rule.set_scope(name, *value) {
+                let res = match rule.set_scope(name, value.clone()) {
```
divarvel, line 164: does this have an impact on performance?

divarvel (commented):
This generally looks good to me. got a few questions and nit-level comments.

biscuit-auth/src/token/authorizer.rs
saoirse-a: This public key is just the algorithm identifier and the binary representation of the key rather than the type from a library which implements `Verify`. Using this representation in authorisation (which just needs to do equality comparison of public keys) avoids needlessly parameterising these by a crypto algorithm.

I aliased the type to `InertPublicKey` anywhere the public API `PublicKey` was also used in the same module.

biscuit-auth/src/token/builder/authorizer.rs
saoirse-a: Not likely much. The reason they're not copy is that the inert form of public key uses a vector rather than a byte array, but they're the same amount of data. It's some additional malloc/frees vs the cost of memcpying somewhat large byte arrays.

The main reason I chose this representation is that I chose `Vec` in `SerializePublicKey` is because it is not possible using const generics to write traits with a "key length" as a constant field. It is possible using typenum and some crypto libraries do this but I am personally pretty strongly against using typenum.

We could possibly have a public "serialised public key" representation which is an enum of supported algorithms carrying the public key data as a fixed length array instead of `(Algorithm, Vec<u8>)` but I don't think it matters that much.

biscuit-auth/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# `7.0.0`
+
+- Abstract biscuits over the crypto implementation (#334)
```
Geal: maybe a "breaking" section indicating what changed and how to migrate

Geal (approved):

biscuit-auth/src/token/builder/biscuit.rs
```diff
@@ -124,34 +125,34 @@ impl BiscuitBuilder {
         f
     }
 
-    pub fn build(self, root_key: &KeyPair) -> Result<Biscuit, error::Token> {
+    pub fn build<K: SerializePrivateKey>(self, root_key: &K) -> Result<Biscuit<K>, error::Token> {
```
wbourne0: If the root key is in a HSM/TPM, it won't be serializable.  Therefore this trait cannot be implemented for a HSM/TPM.

biscuit-auth/src/token/builder/biscuit.rs
```diff
@@ -124,34 +125,34 @@ impl BiscuitBuilder {
...
 
-    pub fn build_with_key_pair(
+    pub fn build_with_key_pair<K: SerializePrivateKey>(
         self,
-        root: &KeyPair,
+        root: &K,
         symbols: SymbolTable,
-        next: &KeyPair,
-    ) -> Result<Biscuit, error::Token> {
+        next: &K,
```
wbourne0: Having the root key and next keys be the same type is a bad idea; it's better to have them be separate.

biscuit-auth/src/token/builder/biscuit.rs
```diff
@@ -124,34 +125,34 @@ impl BiscuitBuilder {
...
         self,
-        root: &KeyPair,
+        root: &K,
         symbols: SymbolTable,
         rng: &mut R,
-    ) -> Result<Biscuit, error::Token> {
+    ) -> Result<Biscuit<K>, error::Token> {
         let authority_block = self.inner.build(symbols.clone());
         Biscuit::new_with_rng(rng, self.root_key_id, root, symbols, authority_block)
     }
```
wbourne0, line 147: This is implying that the root key must not only be the same type as the result key, but that it must also seedable from an RNG.

wbourne0 (changes requested):
I don't fully understand what the point of this is, since it doesn't allow for HSMs or TPMs.  The private keys are still required to be serializable.  Which means we must have the private material outside of it.  

I don't know the codebase that well (first time looking at it today) but I don't think that having a custom public key trait makes much sense here, as there's currently only 2 algorithms. 

I haven't looked through all of the library yet, but I'm also not convinced that this needs to be a breaking change.

I think this should be started off by just add a `Signer` trait and allowing issuance with root keys (a simple switch to `&impl Signer` for the root key reference, as long as the `KeyPair` type `impl` this trait, all previous calls should be compatible).

biscuit-auth/src/crypto/traits.rs
wbourne0: I'd say this goes against the general rust naming conventions.  Usually for a trait to be a verb, the trait represents something where that verb can be applied to.  So an impl of `Hash` for a type allows you to hash that type.

For a trait that represents something that *performs* an action, you usually have a noun - e.g. a `Hasher`.  I don't know if there's a better way of wording this, but there's a discussion I found [here](https://github.com/rust-lang/api-guidelines/discussions/28) which at a glance seems to have some good guidelines / discussion.

biscuit-auth/src/crypto/mod.rs
```diff
@@ -11,71 +11,62 @@
...
+use rand_core::{CryptoRng, RngCore};
+use zeroize::Zeroizing;
+
+use crate::builder::Algorithm;
+use crate::format::schema;
+use crate::format::ThirdPartyVerificationMode;
+
+use super::error;
+
+pub use self::traits::{Verify, Sign, SerializePublicKey, SerializePrivateKey};
```
wbourne0, line 32: These aren't re-exported in `lib.rs`

biscuit-auth/src/token/builder/biscuit.rs
saoirse-a: You're right in this and the other comment, the builder part of the API is not forward compatible with HSM keys even though the split in the traits was intended to support it. I'll revise them.

biscuit-auth/src/crypto/traits.rs
saoirse-a: I have no strong opinion about these names and I'm happy to change them to whatever the project prefers.

biscuit-auth/tests/macros.rs
saoirse-a: Done

biscuit-auth/src/token/public_keys.rs
saoirse-a: I got rid of both `insert_serialize` and `insert_proto` and had the callers do the conversion themselves.

biscuit-auth/src/token/public_keys.rs
saoirse-a: I got rid of both `insert_serialize` and `insert_proto` and had the callers do the conversion themselves.

biscuit-auth/CHANGELOG.md
saoirse-a: Made the changelog update a lot more complete and added a section of breaking changes

biscuit-auth/src/crypto/mod.rs
saoirse-a: Fixed

biscuit-auth/src/token/builder/biscuit.rs
saoirse-a: Fixed

biscuit-auth/src/token/builder/biscuit.rs
saoirse-a: Fixed

biscuit-auth/src/token/builder/biscuit.rs
saoirse-a: Fixed

saoirse-a:
> I don't fully understand what the point of this is, since it doesn't allow for HSMs or TPMs. The private keys are still required to be serializable. Which means we must have the private material outside of it.

The original goal of this work was to support other crypto implementations (not algorithms) if users want to use a different library from p256/ed25519-dalek for either performance or security assurance reasons. That's what your change in #340 doesn't fully support; you could swap the issuer key but all internal signature verifications have to route through p256/ed255190-dalek. p256 in particular is on the order of 5x slower than some other implementations (see benchmarks in #320).

But being forward compatible with HSMs was definitely also a goal of this change, which is why we went this route instead of the route in #320. You spotted that the builder APIs as expressed actually aren't compatible with that, which was a big oversight on my part. To fix that mistake, `BiscuitBuilder` now distinguishes between an `RK` parameter (the root key) and the `K` parameter of the biscuit, which is the proof key.

This introduced a new issue that the builder APIs don't have an obvious way to infer the `K` parameter if its not the same as the root key, since users are generating that key from an RNG and never directly have a value of `K`. I didn't want users to have to turbo fish in a type for `K` if they're using the standard implementations, so I gave `BiscuitBuilder` a phantom `K` parameter which like in `Biscuit` is defaulted to `PrivateKey`.

`BiscuitBuilder::new` constructs a `BiscuitBuilder<PrivateKey>` so users who want to use the default implementation don't have to specify anything. Even if they're using a custom RK like one backed by an HSM, that will be inferred from usage. For users who want to select their own `K` because they don't want to use the provided types, the implementation of `Default` allows any K, so they would write `let builder: BiscuitBuilder<MyPrivateKeyType> = BiscuitBuilder::default()`. This is analogous to how `HashMap::new` and `HashMap::default` work with regard to the hasher type and it seems like the best solution.

---

I've also renamed the `public_keys::PublicKey` to `PublicKeyData` to make it less confusing than having two types called `PublicKey`. I've made its constructors public and added `algorithm` to its API, since it is ultimately a part of the public API via the scope DSL. With some patches to test cases it now builds and passes.

The changelog is also more complete and I've responded to @divarvel's nits. Other than the question of trait naming, if we're happy with this direction I think this PR is ready to merge.

saoirse-a:
One more set of related changes: I wanted to make it so you could pass any public key type (including your own custom ones) to the Datalog macros. In the process I realised that `ToAnyParam` and `AnyParam` aren't actually doing anything anymore: macros already distinguish between scopes and terms. This let me delete a good amount of code from the Datalog builders; now for terms you need to pass something that implements `Into<Term>` and for scopes you need to pass something that implements `Into<PublicKeyData>`. biscuit-quote also changed to generate calls to `set_lenient` methods directly.

biscuit-auth/src/crypto/traits.rs
divarvel: I’m in favor of merging as-is,but staying open to better suggestions before we cut a release.

divarvel:
> One more set of related changes: I wanted to make it so you could pass any public key type (including your own custom ones) to the Datalog macros. In the process I realised that `ToAnyParam` and `AnyParam` aren't actually doing anything anymore: macros already distinguish between scopes and terms. This let me delete a good amount of code from the Datalog builders; now for terms you need to pass something that implements `Into<Term>` and for scopes you need to pass something that implements `Into<PublicKeyData>`. biscuit-quote also changed to generate calls to `set_lenient` methods directly.

Thanks! I actually worked on this in https://github.com/eclipse-biscuit/biscuit-rust/pull/314 but it was stalled because at that time it was a breaking change.

divarvel:
Thanks for the effort @saoirse-a, and thanks for the valuable feedback @wbourne0 

I’ll merge as is, i’m still open to naming improvements before cutting a release, but i think this is fine as it is right now.

This solves two huge pain points so I’m glad to see this move forward.

divarvel:
Could you have a look at the failing coverage test?

wbourne0:
> The original goal of this work was to support other crypto implementations (not algorithms) if users want to use a different library from p256/ed25519-dalek for either performance or security assurance reasons. That's what your change in #340 doesn't fully support; you could swap the issuer key but all internal signature verifications have to route through p256/ed255190-dalek. p256 in particular is on the order of 5x slower than some other implementations (see benchmarks in #320).

Yes, #340 is intended to be a gradual change with no breaking changes, I was intentionally keeping it minimal to avoid locking into specific architectural decisions.  I only added `From` impls for those two types because it was the most minimal way I could think of to somewhat validate that the signatures are actually the correct format (at a type level / without potentially adding overhead).

I also don't see why this needs to be a breaking change again, especially if you're setting default parameters.  Adding traits doesn't need to be a breaking change; though removing them usually is.

As for supporting other implementations of ed25519 and p256, personally I'm of the opinion that other options should be implemented via feature flags and optional dependencies (with current config under the `default` feature).  I think in general with token formats you want to keep the user as far away from it as possible.   It feels more reliable to have a first party integration than it does to provide your own.

This may be addressed but I'm also a little concerned about this implementation if a biscuit has mixed key types; I wonder if the ideal type for the validator type should enforce that validators for both be written (e.g. with associated types). 

Anyhow, this is a large change to add all at once (and I'm not super familiar with the codebase in general) so I have a hard time visually validating it.  I don't own this repo so do as you will, but I do think it'd be good to see some e2e examples of using this before merging it.  It's probably one of the best ways to test a change like this.

peteman-oai:
Thank you folks for pushing this through, this seems like a nice step forward. We also have a need for external signing, similar to @wbourne0 and @echo-ilabs. This PR looks like a step in that direction, but not quite there yet because the API is synchronous and not yet part of a release.

Is anyone tracking any next steps on that front?

PR #338, opened 2026-07-21
Add an external-signer path for third-party blocks (HSM/TPM/Secure Enclave support)

### LINKED ISSUE #339: Third-party blocks cannot be signed by hardware-held keys
ThirdPartyRequest::create_block takes the signer's raw &PrivateKey. Keys held in HSMs, TPMs, Apple's Secure Enclave or Android StrongBox can't be used: those modules will sign bytes for you, but never release the key, and the only workaround today is keeping a software copy -- which defeats the point of hardware-bound keys.

It would be enough to split create_block in two: one call that returns the exact bytes to sign, and one that assembles the block from an externally produced signature.

echo-ilabs:
Closes #339.

`create_block` takes the signer's raw `&PrivateKey`, which rules out keys held in HSMs, TPMs and secure enclaves: they will sign bytes for you, but never release the key. This splits the operation in two, additively:

```rust
let unsigned = biscuit.third_party_request()?.prepare_block(block_builder)?;
let sig: Vec<u8> = hardware.sign(unsigned.bytes_to_sign());   // key never leaves the module
let block = unsigned.with_external_signature(public_key, &sig)?;
biscuit.append_third_party(public_key, block)?;
```

- `prepare_block` does everything `create_block` did up to the signature and exposes the exact bytes to sign.
- `with_external_signature` assembles the block, verifying the signature against the supplied public key up front -- so a wrong key or a corrupted signature fails right there instead of later, as an opaque deserialization error on the recipient's side.
- `create_block` is now built on top of the split path, so the two stay equivalent by construction.

No wire-format or verification changes. Tests cover external-signing round trips for Ed25519 and P-256 (the algorithm hardware modules typically hold, verified by full re-deserialization), byte-for-byte equivalence of the two paths for deterministic Ed25519, and tamper/wrong-key rejection at assembly.

### DIFF
--- a/biscuit-auth/src/lib.rs
+++ b/biscuit-auth/src/lib.rs
@@ -259,7 +259,7 @@ pub use token::builder_ext;
 pub use token::unverified::UnverifiedBiscuit;
 pub use token::Biscuit;
 pub use token::RootKeyProvider;
-pub use token::{ThirdPartyBlock, ThirdPartyRequest};
+pub use token::{ThirdPartyBlock, ThirdPartyRequest, ThirdPartyUnsignedBlock};
 
 #[cfg(feature = "bwk")]
 mod bwk;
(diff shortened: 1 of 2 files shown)


divarvel:
Hi, this is one of the goals of #334 which should be merged soon.

Can you have a look at it and check if that matches your use-case?

Issue #317, opened 2025-10-31
Support FIPS with aws-lc-rs


skeggse:
The `aws-lc-rs` crate supports a `fips` feature flag that links the crate to a FIPS-validated crypto module. It'd be really helpful if `biscuit-auth` supported the use of modular cryptographic crates similar to how `rustls` supports multiple crypto backends!

Along these lines, it'd also be helpful to be able to enable/disable particular cryptographic algorithms using feature flags, to disconnect particular unused dependency crates. For example, if I only use `ecdsa`, I might want to exclude `rsa` as a dependency.

Issue #215, opened 2024-05-09
Proposal: add generic interface for external key registries 


ShockleyJE:
We are evaluating biscuits and the project is extremely appealing for our use case, save for the requirement that the implementing system requires access to the private key itself, which precludes usage with managed key services like AWS KMS. 

I'm not deeply familiar with the project history, future goals, and codebase, but it does look possible technically to add an interface for externalizing the [signing](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#key-usage) & [public key download](https://docs.aws.amazon.com/kms/latest/developerguide/download-public-key.html) responsibilities of `KeyPair` as an optional feature in the crate, where implementations of these responsibilities would be implemented.

The integration with providers would be a non-goal, other than providing a best-effort example.

If this proposal is consistent with the direction of the project we can work around this limitation in the interim and I can offer to contribute to the implementation if desired

divarvel:
Hi, this looks similar to an open PR in biscuit-go https://github.com/biscuit-auth/biscuit-go/pull/130

The token format still requires to have access to the private key for non-authority blocks, but I think that would be okay in this scenario, we only care about delegating signature to a KMS for the authority block (and third-party blocks, but that's similar).

ShockleyJE:
Adding an external `sign` interface would be the _minimal_ solution. The requirement for direct private key access is the main limiting factor of being able to integrate biscuits with modern key management systems, where direct access to the private key is not possible 

I would add a stretch goal to ensure that `verify` functionality, in addition to sign, is able to be externalized from biscuits as well.

Some users will want to use their key management service's [verify](https://docs.rs/aws-sdk-kms/1.22.0/aws_sdk_kms/client/struct.Client.html#method.verify) functionality, and some will want to [cache public keys](https://docs.rs/aws-sdk-kms/1.22.0/aws_sdk_kms/operation/get_public_key/builders/struct.GetPublicKeyFluentBuilder.html) locally rather than making that network request on each verification.

Thanks for the work you do 🙏 No urgency from our side to accommodate this use case- we may return to biscuits for our use case at a later time

PR #340, opened 2026-08-02
Add Signer trait

wbourne0:
Another attempt at allowing for private keys which are not in-memory (e.g. for HSMs, TPMs, KMS).  This is a minimal implementation to add an interface for arbitrary signers.  Currently only supports root keys.  

In the future, an async version of this trait could be added (which would be great for KMS and similar).  For now this is considered out of scope, as async versions of the rest of the code would need to be added.

## Closes
- #317 (external signer half at least - does not add first party aws support)
- #215

The closest other PR to this is #334, but it's adding a lot more at once and seems to be incomplete.    This PR aims to be as simple as possible and to be non-breaking.

#320 may be closed by this too, depending on whether first party AWS support is desired or not.


biscuit-auth/src/crypto/mod.rs
```diff
@@ -429,13 +447,13 @@ impl fmt::Display for PublicKey {
 }
 
 #[derive(Clone, Debug)]
+/// A signature of a [Biscuit](crate::Biscuit) block.
+///
+/// May be constructed via [Into] from  [ed25519_dalek::Signature] or
+/// from [ecdsa::Signature] with the [NistP256](::p256::NistP256) curve.
 pub struct Signature(pub(crate) Vec<u8>);
```
wbourne0, line 455: Previously this was implicitly public as a return type; now it's explicit (hence the added comment) and addressable.

There may be some merit with regards to changing this into an `enum` type, although this may technically be a breaking change.

biscuit-auth/src/crypto/mod.rs
```diff
@@ -429,13 +447,13 @@ impl fmt::Display for PublicKey {
...
 
 impl Signature {
-    pub fn from_bytes(data: &[u8]) -> Result<Self, error::Format> {
-        Ok(Signature(data.to_owned()))
-    }
-
```
wbourne0, line 438: This was never usable before as `Signature` wasn't exposed publicly.  I've opted to remove it since it's fairly unsafe.  Happy to add it back.

biscuit-auth/src/crypto/traits.rs
```diff
@@ -0,0 +1,18 @@
...
+///
+/// Instances of `Signer` may be used with [BiscuitBuilder] as root keys.
+pub trait Signer {
+    /// The algorithm used.  Must match the signature returned via [sign](Self::sign).
+    fn algorithm(&self) -> Algorithm;
+    /// Sign a series of bytes, returning a signature.  This signature must match
+    /// what [self.algorithm()](Self::algorithm) returns.  Any incorrect values
+    /// will likely result in invalid tokens.
+    fn sign(&self, data: &[u8]) -> Result<Signature, error::Format>;
+}
```
wbourne0, line 18: I don't super like this API, but it should do fine.  The other option would be to have `sign` (or the `Signer` trait itself) take an algorithm type as an argument, where that type must impl `Algorithm` or similar.  

Also, this may be better named as `try_sign` to follow the convention used in the `ed25519_dalek` and `ecdsa` crates.

Could also add an associated `Error` type, although then we'd have to pass it back.

biscuit-auth/src/format/mod.rs
```diff
@@ -568,8 +570,8 @@ where
         _ => {}
     }
 
-    match (block_keypair, next_keypair) {
-        (KeyPair::Ed25519(_), KeyPair::Ed25519(_)) => {}
+    match (block_keypair.algorithm(), next_keypair.algorithm()) {
+        (Algorithm::Ed25519, SchemaAlgorithm::Ed25519) => {}
```
wbourne0, line 574: Kinda gross here (the `SchemaAlgorithm`) but I think it's a little neater than relying on the underlying enum structure.

biscuit-auth/src/crypto/traits.rs
wbourne0: Notably we want this separate from a potential later `PrivateKey` trait, as root keys don't actually need to provide public keys.  At most they may optionally have an ID.

